PR #1, opened 2022-11-10
Introduce release action

kadamwhite:
This PR introduces a GitHub action to build generated assets into the `release` branch every time code is merge to `main`. It assumes that `release` branch commits will be tagged for release via Composer or else included in other projects as a submodule.

This is the history of the release branch when using this action, as [tested earlier today in this repository](https://github.com/wikimedia/shiro-wordpress-theme/actions):

```
*   eede1a3e (HEAD -> release, origin/release) Build to 'release' from main#5b2685f8
|\  
| * 5b2685f8 (origin/main, main) Bump build
| * 8e976671 Use shorter ref name, full disambiguation not needed
* | 1672b027 Build to 'release' from main#d75887a1
|\| 
| * d75887a1 Remove test code
* | 91cafb02 Build to 'release' from main#6aede1f4
|\| 
| * 6aede1f4 Change concurrency group to be more general
* | 4853956b Build to 'release' from main#1796000b
|\| 
| * 1796000b Bump
| * 656b86f4 Add concurrency setting
```
Every change to main that gets pushed, `release` gets updated with the latest source branch code and then any generated assets are force-committed.

The `action.yml` file could be extracted into a reusable composite action, but I didn't want to take that step until we were certain we'd need it.

### DIFF
--- a/.github/actions/build-to-release-branch/action.yml
+++ b/.github/actions/build-to-release-branch/action.yml
@@ -1,0 +1,57 @@
+name: Build to release branch
+description: Merge and build a source branch into a releasable, fully-built branch.
+
+inputs:
+  source_branch:
+    description: Name of source branch.
+    required: true
+  release_branch:
+    description: Name of target release branch. Will be updated with a merge commit including built code.
+    required: true
+  built_asset_paths:
+    description: Generated assets to include in release branch commit. Overrides .gitignore.
+    required: true
+  build_script:
+    description: Command to run to build code. "npm run build" by default.
+    required: false
+
+runs:
+  using: composite
+
+  steps:
+    - name: Configure git
+      shell: bash
+      run: |
+        git config user.name "Your friendly neighborhood GH Actions Bot"
+        git config user.email "<>"
+
+    - name: Check out release branch
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ inputs.release_branch }}
+        fetch-depth: 0
+
+    - name: Merge source into release branch
+      shell: bash
+      # Simulate "-s theirs", which doesn't exist in git for reasons.
+      # https://web.archive.org/web/20131126171744/http://www.seanius.net/blog/2011/02/git-merge-s-theirs/
+      run: |
+        git fetch origin ${{ inputs.source_branch }}
+        git fetch origin ${{ inputs.release_branch }}
+        git checkout ${{ inputs.source_branch }}
+        git checkout ${{ inputs.release_branch }}
+        git merge -s ours ${{ inputs.source_branch }}
+        git diff --binary ${{ inputs.source_branch }} | git apply -R --index
+        git commit --amend -m "Merge branch '${{ inputs.source_branch }}' into '${{ inputs.release_branch }}'"
+    
+    - name: Build code
+      shell: bash
+      run: ${{ inputs.build_script }}
+
+    - name: Update release branch
+      shell: bash
+      run: |
+        echo "${{ inputs.built_asset_paths }}"
+        git add -f ${{ inputs.built_asset_paths }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -2,34 +2,31 @@ name: Release
...
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          ORIGINAL_TAG_PREFIX: original/
-          OUTPUT_BUILD_INFO_FILENAME: build.json
-          TEST_TAG_PREFIX: test/
+          source_branch: main
+          release_branch: release
+          built_asset_paths: ./*.css package-lock.json map assets/dist
+          build_script: |
+            npm ci
+            npm run build:gulp
```
kadamwhite, line 32: This will need to be expanded to include the webpack build, but that crashes at present. So, this was the MVP for proving that the approach works.

.github/actions/build-to-release-branch/action.yml
```diff
@@ -0,0 +1,58 @@
...
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ inputs.release_branch }}
+        fetch-depth: 0
+
+    - name: Merge source into release branch
+      shell: bash
+      # Simulate "-s theirs", which doesn't exist in git for sensible reasons
+      # which do not apply to our current situation.
+      # https://web.archive.org/web/20131126171744/http://www.seanius.net/blog/2011/02/git-merge-s-theirs/
```
kadamwhite, line 38: Sadly this excellent post now exists only in the web archive, and by reference on SO. Entropy, it's a thing.

.github/actions/build-to-release-branch/action.yml
```diff
@@ -0,0 +1,58 @@
...
+      run: |
+        git fetch origin ${{ inputs.source_branch }}
+        git fetch origin ${{ inputs.release_branch }}
```
kadamwhite, line 41: We should not, strictly, need to fetch the source branch. We check it out above, after all. But it felt thorough to do so and doesn't seem to increase the run time significantly.

.github/workflows/release.yml
```diff
@@ -2,34 +2,31 @@ name: Release
...
 on:
   push:
-    tags:
-      - '*'
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
```
goldenapples, line 10: ~I get why you'd want to cancel in-progress merges, but is there a chance that this would break the order of history entries?~ never mind, thinking about this it seems like it should be fine. In the worst case, the release branch might not have the most recent code from main, but it'll be fixed with the next push.

goldenapples (approved):
This is really nice. I like how clean the history looks on these merges and it's really easy to get from a built commit back to the corresponding history for bisecting etc.

Only concern I have is that this might make it *too* easy for someone to accidentally start a branch off of release and end up checking built assets back in to main. I suppose that'd get caught in code review though so there's probably no need to protect against it technically...

.github/workflows/release.yml
kadamwhite: > concurrency

My goal is more to avoid any potential race condition if one build took longer than the later one, and consider it more important that `release` have the most recent build over all other concerns. I think if two commits are merged in quick succession, there's usually little benefit to generating intermediate builds. Open to disagreement though!

kadamwhite: > `fetch_depth: 0`

Because I also specify `ref: ${{ inputs.release_branch }}`, the checkout action will not pull all other branches. (I didn't feel we'd need them).

kadamwhite:
Thank you for the review, @goldenapples! I agree we'd want to avoid checking built assets in, but figure we can cross that bridge when we get there—perhaps a CI action for PR branches that ensures there are no files within certain paths.

PR #2, opened 2022-11-10
Add release branch protection workflow

kadamwhite:
This workflow iterates on a suggestion from @goldenapples in #1 that it would be bad for the release branch to be merged back into other branches. While it is not possible to prevent a branch from being used as a merge source in GitHub, this action should fail the PR if such a merge is attempted.

What it looks like if the action fails:
image

### DIFF
--- a/.github/workflows/protect-release-branch.yml
+++ b/.github/workflows/protect-release-branch.yml
@@ -1,0 +1,15 @@
+name: Branch protection
+on:
+  pull_request:
+    branches:
+      - 'release'
+
+jobs:
+  release:
+    name: Do not merge the Release branch into other branches
+    runs-on: ubuntu-latest
+    steps:
+      - shell: bash
+        run: |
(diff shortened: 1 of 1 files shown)


goldenapples (commented):
I don't think this will do what I wanted it to. This will automatically fail pull requests that are targeting the `release` branch, but the bigger concern I would have is if a branch was initially started from the `release` branch, and then opened as a pull request into `main`.

The only concern I had is that there are [built asset files](https://github.com/wikimedia/shiro-wordpress-theme/pull/1/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R29) added in the release script using `git add -f`, and they'd get added into the main branch where they're git ignores and wouldn't be updated. This is not a big problem at all but it would make history look messier than ideal.

The only thing I could think of is to maintain the list of built assets paths that are added using git add --force in the release script and making sure that none of them are present in a PR.

PR #3, opened 2022-11-11
Reconcile versions of theme

kadamwhite:
This merges the full tree history of the `shiro` theme as it exists on the Foundation site, into this separate repository. I extracted it using `git filter-repo --subdirectory-filter themes/shiro`.


kadamwhite:
The only thing that gives me pause here in the diff is the removal of `template-report.php`. Anybody able to confirm that one is good to be removed? @goldenapples , @smccafferty ...?

kadamwhite:
That partial got removed in https://github.com/humanmade/Wikimedia/pull/570/files and isn't referenced anywhere I can see. Think the removal of `template-report.php` is 🆗

PR #4, opened 2022-11-11
685- Added <em> tag as allowed

pamprn09:
I've added `em` tag into wp_kses allowed html array.

dashboard

em-tag

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -527,7 +527,7 @@ function shiro_safe_title( string $title ): void {
 	echo wp_kses(
 		$title,
 		[
-			'span' => ['class'],
+			'span' => [ 'class' ],
 			'em' => [],
 		]
 	);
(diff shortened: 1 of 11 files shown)


functions.php
```diff
@@ -502,3 +502,19 @@ function shiro_add_slug_body_class( $classes ) {
...
 }
 add_filter( 'body_class', 'shiro_add_slug_body_class' );
+
+/**
+ * Allow span and em tags into titles.
+ *
+ * @param string $title Post title.
+ * @return string
+ */
+function the_safe_title( string $title ): void {
```
alwaysblank, line 512: I'd consider putting this in `inc/template-functions.php` or `inc/template-tags.php` and changing the name to `wmf_safe_title()` or something similar for consistency.

kadamwhite (changes requested):
+1 to Ben's suggestion -- and I think there may be another couple spots we didn't know about that should also be updated for consistency. Searching for `esc_html( $title )` for example, I see a bunch of results:

image

Not all of those may be post titles, but could you look around the theme code and introduce this function in other places where we're echoing or rendering a post title?

kadamwhite (approved):

PR #5, opened 2022-11-11
Fix JS lint errors blocking build, and upgrade toolchain

kadamwhite:
This does a bunch, but it's in several key buckets

- Fix code style errors that blocked or showed errors during build locally
  - 90% of changes were applied with `npm run lint -- --fix`
- Upgrade the build to use the latest version of Webpack and Webpack Helpers
  - Fixes a _wealth_ of SASS warnings shown on build, including the deprecation of division—I used an auto-migrator script to apply that change.
- Extend linting to build code like the webpack configuration
- Compress large images in the `assets` directory using [`jpegoptim`](https://github.com/tjko/jpegoptim) for visually-lossless compression


goldenapples (approved):
It all looks sensical on initial readthrough, and the build passes, so I'd say :+1:.

PR #8, opened 2022-11-15
Address PHPCS errors in files touched by #4

kadamwhite:
Now that we have PHP linting wired up, some PRs will trigger an onslaught of PHPCS errors. This PR preemptively addresses  errors in the code touched by #4 so that the feature PR can be addressed independently.

It also resolves a global issue with #7 where outdated versions of the WP standards were being used which could cause fatal errors when running `phpcs` in the GitHub action.

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -61,7 +61,7 @@
 	</rule>
 	<rule ref="WordPress.NamingConventions.PrefixAllGlobals">
 		<properties>
-			<property name="prefixes" type="array" value="wmf,rkv"/>
+			<property name="prefixes" type="array" value="wmf,shiro"/>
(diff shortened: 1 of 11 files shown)


phpcs.xml
```diff
@@ -61,7 +61,7 @@
 	</rule>
 	<rule ref="WordPress.NamingConventions.PrefixAllGlobals">
 		<properties>
-			<property name="prefixes" type="array" value="wmf,rkv"/>
+			<property name="prefixes" type="array" value="wmf,shiro"/>
```
kadamwhite, line 64: We should standardize on `shiro` longer term, but not necessary today.

PR #11, opened 2023-01-23
Support vegalite plugin

kadamwhite:
Permits blocks from [`wikimedia/vegalite-wordpress-plugin`](wikimedia/vegalite-wordpress-plugin) to be rendered in Shiro


kadamwhite:
We should fix the code style errors but I'd like to deem that out of scope for this PR.

pamprn09 (approved):
There's a few corrections to be made, pointed by GH Actions. Other than thta, looks good!

PR #14, opened 2023-01-25
Address PHP warnings observed in VIP logs

kadamwhite:
Several PHP warnings have been reported in the deployed environment, which began to show up after the upgrade to PHP 8 (at which point the undefined array key became a warning). This PR addresses several pervasive warnings, and fixing them will make it easier to recognize true issues in the application health logs.


pamprn09 (approved):
Looks good!

PR #15, opened 2023-01-30
Guard against empty or null values causing errors in production

kadamwhite:
Follow-up to #14 to fix additional errors, including one fatal.


missjwo (approved):
I personally dislike the shorthand syntax, but it works.

alwaysblank (approved):

goldenapples (approved):

PR #16, opened 2023-01-31
530 - Add wikimedia commons specific linkpromo to photo credits

kadamwhite:
Extracted from the old monorepo, copying over for merge and test

From @pamprn09's original PR:

> I added an icon when the image is from Wikimedia Commons. When hovered or in focus, the icon will display a text promoting the utilization of Wikimedia Commons.
>
> Wikimedia-Foundation (2)


kadamwhite:
@pamprn09 Linting is currently broken, which we can look into separately. Can you re-review and see if this has the right logic? It's possible some commits got missed from https://github.com/humanmade/Wikimedia/pull/720

pamprn09 (approved):
Confirming that this should be moved to shiro. Thanks!

kadamwhite:
Opened #21 to address the PHPCS warnings, so that merging this doesn't break the build

PR #17, opened 2023-01-31
426 - Add page id parameter to donation button

kadamwhite:
Copied from @pamprn09's original PR in the old monorepo.

> This PR adds a page_id variable and rewrites Donation buttons URLs.

### DIFF
--- a/assets/src/scripts/modules/donate-buttons.js
+++ b/assets/src/scripts/modules/donate-buttons.js
@@ -1,12 +1,8 @@
 /**
  * Collect any "Donate" buttons on the page and set their utm_source.
- */
-
-/**
+ *
  * Searches for any links to donate.wikimedia.org, and replaces their hrefs to
  * include UTM parameters in the query strings.
- *
- * @returns {HTMLElement[]} Array of donation buttons.
  */
 const init = () => {
 	const donationButtons = [ ...document.querySelectorAll( '[href^="https://donate.wikimedia.org"]' ) ];


pamprn09 (approved):
We do need those on shiro! Thanks for moving this!

assets/src/js/global.js
```diff
@@ -1,3 +1,19 @@
...
+/* eslint-disable */
+
 /**
  * Generic site JavaScript.
  */
+
+/**
+ * Search for Donation Button and add parameter link Op1.
+ */
+document.querySelectorAll('[href^="https://donate.wikimedia.org"]').forEach( function( item ) {
```
kadamwhite, line 11: Now that the theme is in its own repo we should revisit how to integrate this into the main JS bundle, possibly by moving it to `src/scripts/modules` and including it from `scripts/shiro.js`

assets/src/js/global.js
```diff
@@ -1,3 +1,19 @@
...
+	'use strict';
+	var page_id = window.post_id;
+	var url = item.href;
+	var params = new URLSearchParams( url.search );
+	params.set( 'utm_source', page_id );
+	item.href = url.replace( /\?.*/, '?utm_source=' + page_id );
```
kadamwhite, line 17: Probably want 
```suggestion
	item.href = url.replace( /\?.*/, params.toString() );
```
here, to not lose other parameters

functions.php
```diff
@@ -492,3 +492,11 @@ function shiro_add_slug_body_class( $classes ) {
 	return $classes;
 }
 add_filter( 'body_class', 'shiro_add_slug_body_class' );
+
+function shiro_page_id() {
+	global $post;
+	wp_enqueue_script( 'shiro-page-id', 'assets/src/js/global.js' );
+	wp_localize_script( 'shiro-page-id', 'post_id', $post->ID );
```
kadamwhite, line 499: PHPCS is warning about not enqueing these in the footer, which won't be an issue if we can fold it in to the main shiro.js script bundle

kadamwhite (commented):
Capturing a few things we should clean up before shipping this

goldenapples:
Isn't the page ID already being appended to that Donate button in PHP? It looks as though this is already set in the donate.php template: https://github.com/wikimedia/shiro-wordpress-theme/blob/main/template-parts/site-header/donate.php#L16

assets/src/js/global.js
goldenapples: Updated in [9bb40a2](https://github.com/wikimedia/shiro-wordpress-theme/pull/17/commits/9bb40a25747d9e40149eb8bd5825aeafccda3636).

assets/src/js/global.js
goldenapples: Updated in [9bb40a2](https://github.com/wikimedia/shiro-wordpress-theme/pull/17/commits/9bb40a25747d9e40149eb8bd5825aeafccda3636).

varnent:
> Isn't the page ID already being appended to that Donate button in PHP? It looks as though this is already set in the donate.php template: https://github.com/wikimedia/shiro-wordpress-theme/blob/main/template-parts/site-header/donate.php#L16

I will look into this and get back to you.

varnent:
> Isn't the page ID already being appended to that Donate button in PHP? It looks as though this is already set in the donate.php template: https://github.com/wikimedia/shiro-wordpress-theme/blob/main/template-parts/site-header/donate.php#L16

Looking into this - this is the only place where they are working - so we may need to remove that duplicate, but should proceed with testing this.

MiguelAxcar (approved):
Minor comments only, it looks good to me :+1:

MiguelAxcar (approved):
Minor comments only, it looks good to me :+1:

assets/src/scripts/modules/donate-buttons.js
```diff
@@ -0,0 +1,27 @@
...
+const donationButtons = [ ...document.querySelectorAll( '[href^="https://donate.wikimedia.org"]' ) ];
+const { object_id } = window.shiro;
+
+/**
+ * Searches for any links to donate.wikimedia.org, and replaces their hrefs to
+ * include UTM parameters in the query strings.
+ *
+ * @returns {HTMLElement[]} Array of donation buttons.
+ */
+const init = () => donationButtons.forEach(
```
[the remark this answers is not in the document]
kadamwhite: Because we're spreading the qsa results into an array, it will always be at worst `[]` -- which does have a forEach method, so it would execute safely even if there are no buttons. You can test this yourself with

```js
[ ...document.querySelectorAll( '.does-not-exist' ) ]
    .forEach( () => console.log( 'Should not run' ) );
```
which should execute without error.

In most modern browsers, the qsa result itself (which is a [NodeList](https://developer.mozilla.org/en-US/docs/Web/API/NodeList)) has a `forEach` method too—so this should also work safely:

```js
document.querySelectorAll( '.does-not-exist' )
    .forEach( () => console.log( 'Should not run' ) );
```

The array spread is a safeguard for compatibility which we _may_ not need, but it's still useful in my opinion. It's easier to think about "what can arrays do" than "what can a NodeList do".

kadamwhite:
@MiguelAxcar I've pushed some small changes, and left a note about one of your concerns. I don't think we need to account for `http` since this is a link to a separate donations site that isn't part of this WP instance, so it should never be using an `http` address, even locally.

Ready for final review. @varnent this passes my own testing, but since it's my PR I can't approve it :)

PR #18, opened 2023-01-31
Dynamically patch phpcs to fix vsprint exception

alwaysblank:
PHP 8 environments cause a fatal error when running wpcs PHPCS sniffs:

```
PHP Fatal error:  Uncaught TypeError: vsprintf(): Argument #2 ($values) must be of type array, string given in /Users/akshitsethi/wp-local-docker-sites/snowsoftware-test/wordpress/wp-content/vendor/squizlabs/php_codesniffer/src/Files/File.php:1056
```
[There is a fix for this](https://github.com/WordPress/WordPress-Coding-Standards/commit/7cd46bed1e6a7a2af3fe24c7f4a044da3076d8f4) but it hasn't yet been released for the official repo. Unfortunately it's not as simple as pointing our `composer.json` at the branch containing this updated code because wpcs is a dependency of several of _our_ dependencies, and composer will choke when asked to resolve two different version specifications.

This uses https://github.com/cweagans/composer-patches to patch in the require fix (which is minimal) when composer dependencies are installed via the `patch` system command, resolving the issue.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
     "config": {
         "allow-plugins": {
             "composer/installers": true,
-            "dealerdirect/phpcodesniffer-composer-installer": true
+            "dealerdirect/phpcodesniffer-composer-installer": true,
+            "cweagans/composer-patches": true
         }
     },
     "require-dev": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -29,5 +30,15 @@
         "phpcs": "phpcs .",
         "phpcs:changed": "phpcs $(git diff --name-only --diff-filter=ACMRTUXB $(git merge-base head main) | grep '\\.php')",
         "phpcbf:changed": "phpcbf $(git diff --name-only --diff-filter=ACMRTUXB $(git merge-base head main) | grep '\\.php')"
+    },
+    "require": {
+        "cweagans/composer-patches": "^1.7"
```
goldenapples, line 35: Can this be a dev dependency?

kadamwhite (approved):
@alwaysblank this is a very workable, clean solution! Thank you very much for the proposal

PR #20, opened 2023-02-03
Make Shiro allowed blocks filterable rather than coupling list to theme releases

kadamwhite:
A simple mu-plugin can serve as the compatibility layer for Shiro and other blocks on a per-site basis, rather than requiring every plugin to be known to the theme. This permits site-specific mu-plugins to handle integration.

- [x] Update to use `allowed_block_types_all` instead of deprecated `allowed_block_types`


pamprn09 (approved):
Looks good!

PR #23, opened 2023-02-03
762: Do not add our own custom filter when we can use core's

kadamwhite:
In retrospect, before this goes live, we should avoid introducing a new filter for specifying allowed blocks when we can simply add a higher-priority filter function on core's same `allowed_block_types_all` hook.


pamprn09 (approved):
Looks good!

PR #24, opened 2023-02-07
Fix language tax update when saving content in the block editor

goldenapples:
When creating content locally with the latest version of shiro theme, any attempts to update newly-created posts break with the following error:

    Fatal error: Uncaught Error: Call to undefined function wp_create_term()
    in /srv/www/wikimediafoundation/public_html/wp-content/themes/shiro/inc/taxonomy-content-language.php on line 80

This is happening because in the REST API endpoint hit to update content in the block editor, wp-admin/includes/taxonomy.php isn't loaded by default. This change ensures that this file is required whenever trying to update language taxonomy terms.

### DIFF
--- a/inc/taxonomy-content-language.php
+++ b/inc/taxonomy-content-language.php
@@ -85,23 +85,21 @@ function wmf_get_and_maybe_create_current_language_term(): ?WP_Term {
  */
 function wmf_create_current_language_term(): ?WP_Term {
 
-	// When this function is called from the REST API, the admin taxonomy
-	// utilities may not be available. This fix ensures that the content
-	// language functionality works in the block editor.
-	if ( ! function_exists( 'wp_create_term' ) ) {
-		require_once ABSPATH . 'wp-admin/includes/taxonomy.php';
-	}
-
 	// Make sure that we always have the "main language" term.
 	$main_locale = wmf_get_current_content_language_term();
-	if ( ! is_a( $main_locale, WP_Term::class ) ) {
-		$result = wp_create_term( wmf_get_current_content_language_slug(), 'content-language' );
-		if ( is_wp_error( $result ) ) {
-			return null;
-		}
-		return get_term( $result['term_id'], 'content-language' );
+
+	if ( is_a( $main_locale, WP_Term::class ) ) {
+		return $main_locale;
 	}
-	return $main_locale;
+
+	// If the current language term doesn't exist, create it now and return it.
+	$created_term = wp_insert_term( wmf_get_current_content_language_slug(), 'content-language' );
+
+	if ( is_wp_error( $created_term ) ) {
+		return null;
+	}
+
+	return get_term( $created_term['term_id'] , 'content-language' );
 }
 
 /**
--- a/inc/taxonomy-content-language.php
+++ b/inc/taxonomy-content-language.php
@@ -1,4 +1,14 @@
 <?php
+/**
+ * "Content Language" taxonomy functions.
+ *
+ * In order to allow creating content in languages that don't have a dedicated
+ * site, we register "Content Language" as a taxonomy which can be set on a
+ * per-post basis. This allows for individual posts relevant to communities not
+ * otherwise served by the blog to be translated and associated with the
+ * original post in the same way that MultiLingual Press handles associating
+ * content between different lnaguage sites.
+ */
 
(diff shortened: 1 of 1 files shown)


inc/taxonomy-content-language.php
```diff
@@ -9,7 +9,7 @@
  */
 function wmf_register_content_language_taxonomy(): void {
 	$language_type_args = array(
-		'heirarchical' => false,
+		'hierarchical' => false,
```
goldenapples, line 12: Unrelated, but the typo bugged me :)

This doesn't change any behavior, because the default for 'hierarchical' is already false.

inc/taxonomy-content-language.php
```diff
@@ -74,6 +74,11 @@ function wmf_get_and_maybe_create_current_language_term(): ?WP_Term {
  * @return \WP_Term|null
  */
 function wmf_create_current_language_term(): ?WP_Term {
+	// When this function is called from the REST API, the admin taxonomy
+	// utilities may not be available. This fix ensures that the content
+	// language functionality works in the block editor.
+	require_once ABSPATH . 'wp-admin/includes/taxonomy.php';
```
kadamwhite, line 80: Should this be gated in an `if ( ! function_exists() )`? I know it'll work either way, but not aware whether there's any theoretical downside to calling this when not needed

kadamwhite (commented):
I'd suggest adding a commit to resolve the PHPCS errors flagged here, we found that it sometimes caused issues in the build actions if the PR was merged with known lint errors.

kadamwhite:
@goldenapples I chatted with @rmccue about this and we both feel this may be a bug in core, if the taxonomy file isn't loaded already.

inc/taxonomy-content-language.php
goldenapples: Makes sense. Intuitively I think of `require_once` as including that same check, but being explicit adds clarity.

goldenapples:
I've fixed the linting issues in this file; ready for re-review.

inc/taxonomy-content-language.php
```diff
@@ -74,6 +84,14 @@ function wmf_get_and_maybe_create_current_language_term(): ?WP_Term {
  * @return \WP_Term|null
  */
 function wmf_create_current_language_term(): ?WP_Term {
+
+	// When this function is called from the REST API, the admin taxonomy
+	// utilities may not be available. This fix ensures that the content
+	// language functionality works in the block editor.
+	if ( ! function_exists( 'wp_create_term' ) ) {
+		require_once ABSPATH . 'wp-admin/includes/taxonomy.php';
```
kadamwhite, line 92: Wouldn't it be cleaner to simply use `wp_insert_term`, and check for term existence if needed, instead of loading all the rest of the admin-specific code?

This is the implementation of wp_create_term: 

```php
function wp_create_term( $tag_name, $taxonomy = 'post_tag' ) {
	$id = term_exists( $tag_name, $taxonomy );
	if ( $id ) {
		return $id;
	}

	return wp_insert_term( $tag_name, $taxonomy );
}
```

The logic could be adjusted to,

```php
	if ( ! is_a( $main_locale, WP_Term::class ) ) {
		$term_id = term_exists( $tag_name, $taxonomy );
		if ( empty( $term_id ) ) {
			$result = wp_insert_term( wmf_get_current_content_language_slug(), 'content-language' );
			if ( is_wp_error( $result ) ) {
				return null;
			}
			$term_id = $result['term_id'];
		}
		return get_term( $term_id, 'content-language' );		
	}
```

and we wouldn't need anything from the admin taxonomy.php.

kadamwhite (changes requested):
Proposal to avoid requiring the admin namespace at all

inc/taxonomy-content-language.php
goldenapples: Yeah, and actually we don't even need the `term_exists` check because that's already handled inside `wmf_get_current_content_language_term()`: https://github.com/wikimedia/shiro-wordpress-theme/blob/7df2deaf3a9faf354dc537f41d6ded11545e1e2a/inc/taxonomy-content-language.php#L57-L60

kadamwhite (approved):
Much cleaner, thanks for inverting the conditionals while doing this!

PR #26, opened 2023-02-08
[hotfix] Fix eslint warning in collapsible-block js handler

goldenapples:
This script is throwing warnings on every build. Adding an explicit return should fix it.


pamprn09 (approved):
Thanks for this fixing! It's coming from my latest PR

PR #28, opened 2023-02-08
#422: Add "home-page-hero-rotator" block

goldenapples:
Adds a wrapper block which can accept any number of "home-page-hero" blocks as children, and update the frontend functionality to rotate between the different hero images, if more than one exist.

See https://github.com/humanmade/wikimedia/issues/422.

Left to do:

- [x] Add block transform to convert an old Home Hero into the new rotator block, so than multiple images can be selected.
- [x] Add editor styling to these blocks, currently they're both a real mess in the editor, and not very intuitive to use at all.
- [x] Refactor frontend rotation code for performance and UX / look and feel.


assets/src/editor/blocks/home-page-hero/style.scss
```diff
@@ -1,5 +1,8 @@
 .editor-styles-wrapper.editor-styles-wrapper {
 	.hero-home {
+		max-width: none !important;
```
pamprn09, line 3: Is there another way to set max-width to none, avoiding the `!important`?

pamprn09 (approved):
This was tested locally and it's working fine. Good job!

goldenapples:
I'm closing out this PR, since the goal is to create a slider block that can eventually be used elsewhere. #32 is a start at registering a more general-purpose carousel block.

goldenapples:
I've also reverted this code out of `develop` to keep the development branch in line with the main branch.

PR #29, opened 2023-02-10
758 - Contrast link

pamprn09:
This is part of ticket [758 - Adding Colors](https://app.zenhub.com/workspaces/wikimedia-foundation-602350f5e5f555000e33e28c/issues/gh/humanmade/wikimedia/758) 

### Add color styles to cards
ezgif com-resize
ezgif com-resize (1)

### Add the white color link when the background has darker colors
**Back**
link-bg-editor
**Front**
link-bg

### DIFF
--- a/assets/src/editor/blocks/card/index.js
+++ b/assets/src/editor/blocks/card/index.js
@@ -12,13 +12,15 @@ import { __ } from '@wordpress/i18n';
 import BlockIcon from '../../../svg/blocks/card.svg';
 import CallToActionPicker from '../../components/cta';
 import ImagePicker from '../../components/image-picker';
+import sharedStyles from '../../helpers/block-styles';
 
 /**
  * Internal dependencies
  */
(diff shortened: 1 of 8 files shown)


assets/src/editor/blocks/card/index.js
```diff
@@ -12,13 +12,15 @@ import { __ } from '@wordpress/i18n';
...
+import sharedStyles from '../../helpers/block-styles';
 
 /**
  * Internal dependencies
  */
 import './style.scss';
 
-export const name = 'shiro/card';
+export const name = 'shiro/card',
+	styles = sharedStyles;
```
kadamwhite, line 23: Nit: Slight stylistic preference for making these two separate statements, rather than using a comma. Comma-separated values seem to be less common with let and const (and export) than they used to be with var.

kadamwhite (approved):

PR #30, opened 2023-03-02
787: Staff Profile Tagging

ajvillegas:
This change allows for multiple member profile team tags to display as comma-separated list of links in the front-end. Single tags display next to the role title and are also outputted as a link.

_Multiple tags:_
Screen Shot 2023-03-02 at 7 04 27 PM

_Single tag:_
Screen Shot 2023-03-02 at 7 05 20 PM

For issue https://github.com/humanmade/wikimedia/issues/787.

### DIFF
--- a/single-profile.php
+++ b/single-profile.php
@@ -17,16 +17,16 @@ while ( have_posts() ) :
 	?>
 
 	<?php
-	$role            = get_the_terms( get_the_ID(), 'role' );
+	$roles           = get_the_terms( get_the_ID(), 'role' );
 	$default_heading = get_theme_mod( 'wmf_related_profiles_heading', __( 'Other members of ', 'shiro-admin' ) );
 	$team_name       = '';
-	$parent_name     = $role[0]->name;
-	$parent_link     = get_term_link( $role[0] );
+	$parent_name     = $roles[0]->name;
+	$parent_link     = get_term_link( $roles[0] );
     $connected_user = get_post_meta( get_the_ID(), 'connected_user', true );
 
-	if ( ! empty( $role ) && ! is_wp_error( $role ) ) {
-		$team_name = $role[0]->name;
-		$ancestors = get_ancestors( $role[0]->term_id, 'role' );
+	if ( ! empty( $roles ) && ! is_wp_error( $roles ) ) {
+		$team_name = $roles;
+		$ancestors = get_ancestors( $roles[0]->term_id, 'role' );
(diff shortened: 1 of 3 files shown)


single-profile.php
```diff
@@ -17,16 +17,16 @@
...
-    $connected_user = get_post_meta( get_the_ID(), 'connected_user', true );
+	$parent_name     = $roles[0]->name;
+	$parent_link     = get_term_link( $roles[0] );
+	$connected_user = get_post_meta( get_the_ID(), 'connected_user', true );
 
-	if ( ! empty( $role ) && ! is_wp_error( $role ) ) {
-		$team_name = $role[0]->name;
-		$ancestors = get_ancestors( $role[0]->term_id, 'role' );
+	if ( ! empty( $roles ) && ! is_wp_error( $roles ) ) {
+		$team_name = $roles;
```
goldenapples, line 28: ~This variable assignment doesn't seem necessary. Are we using it anywhere?~

Never mind, I see it being passed into the `header/profile-single` template part. 

As a purist I'm bothered by the variable name being single when it's expected to be passed as an array, but that ultimately doesn't matter at all :)

single-profile.php
```diff
@@ -66,50 +66,49 @@
...
-                    ?>
-                    <div class="link-list mar-right">
-                    <div class="bold profile-contacts"><a href="/news/author/<?php echo esc_attr( $authorlink ); ?>">
-							<img src="<?php echo esc_url($authorimg); ?>" alt="">
-							<?php echo esc_html( $authorlinkcopy ); ?>
-						</a></div>
-                    </div>
-				<?php endif; ?>
+						<div class="link-list mar-right">
+						<div class="bold profile-contacts"><a href="/news/author/<?php echo esc_attr( $authorlink ); ?>">
```
goldenapples, line 106: Strange that this isn't using get_author_posts_link. But it looks like this duplicates the existing functionality, so it should be fine.

template-parts/header/profile-single.php
```diff
@@ -27,7 +41,13 @@
 		<h1><?php the_title(); ?></h1>
 
 		<p class="post-meta">
-			<?php echo esc_html( $role_desc ); ?>
+			<?php echo wp_kses_post( $role_desc ); ?>
```
goldenapples, line 44: :+1:

goldenapples (approved):
This looks good to me. I can merge it onto develop for client testing.

PR #32, opened 2023-03-06
[#422] Add carousel block

goldenapples:
Registers a "Carousel" block which initializes a basic slide using Splide.js. Allows users to select the type of block to be inserted as each slide (limited to [these options](https://github.com/humanmade/wikimedia/issues/422#issuecomment-1460321073)), the number of slides to display per page, and whether to use navigation arrows, pagination dots, and/or autoplay.


MiguelAxcar (approved):
Although I could not view it rendered on the preview or after updating the page, which is probably expected, I tested it locally and the overall functionality looks well, things are getting together.

assets/src/editor/components/inner-block-slider/index.js
```diff
@@ -0,0 +1,105 @@
...
+import { useSelect, useDispatch } from '@wordpress/data';
+
+import InnerBlocksDisplaySingle from './inner-block-single-display';
+import Navigation from './navigation';
+
+/**
+ * InnerBlockSlider component.
+ *
+ * @param {object} root0 - Component props.
+ * @param {string} root0.parentBlockId - Parent block clientId.
```
kadamwhite, line 14: Nitpick, rename `root0` to `props` for clarity

assets/src/editor/components/inner-block-slider/inner-block-single-display.js
```diff
@@ -0,0 +1,75 @@
+import PropTypes from 'prop-types';
+import { ReactNode, useRef, useEffect } from 'react';
+
+import { useInnerBlocksProps } from '@wordpress/block-editor';
+
+/**
+ * InnerBlockSlider component.
+ *
+ *
+ * @param {object} root0 Component props.
```
kadamwhite, line 10: Same as above

assets/src/sass/blocks/_carousel.scss
```diff
@@ -0,0 +1,79 @@
+@import "../../../../node_modules/@splidejs/splide/src/css/core";
```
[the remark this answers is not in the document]
goldenapples: Nah, the theme styles here are still being built in gulp, and there's no import path handler for node modules. (As an aside, it'd be really nice to rewrite the old gulp build tasks so that at least the JS and CSS parts of it are handled in webpack like the newer parts of the theme codebase...)

assets/src/editor/blocks/carousel/index.js
```diff
@@ -0,0 +1,214 @@
...
+		} );
+
+		// Build options for currentBlock select controller.
+		const allBlocksAvailable = getBlockTypes();
+
+		const blockTypeOptions = ALLOWED_BLOCKS
+			.map( ( blockName ) => {
+				const registeredBlock = allBlocksAvailable.find( ( block ) => block.name === blockName );
+
+				if ( registeredBlock !== 'undefined' ) {
```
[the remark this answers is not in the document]
goldenapples: Ah yeah, I didn't mean to quote it!

assets/src/sass/utilities/_theme-supports.scss
```diff
@@ -14,33 +14,33 @@ $color-palette:
 	blue            $wmui-color-blue            $wmui-color-base100 $wmui-color-blue90,
 	blue-50         $wmui-color-blue50          $wmui-color-base100 $wmui-color-blue90,
 	blue-90         $wmui-color-blue90          $wmui-color-base0   $wmui-color-blue50,
-	blue70          $wmui-color-blue70          $wmui-color-base0   $wmui-color-base50,
+	blue-70          $wmui-color-blue70          $wmui-color-base0   $wmui-color-base50,
```
goldenapples, line 17: @pamprn09 Would changing these color variable names break styling of existing blocks on the site? If so, is there a way to check if these theme colors are being used anywhere on the site?

goldenapples (commented):
The changes here look good to me, I just had a comment about changing the variables that are used to make block style names. It's good to be consistent, but it seems like if any blocks are using the old names, they'll lose their styling after this change. If we make a breaking change like this, we should have a way of updating content after deploying it so we don't break styling. That (could just be doing a search-replace through WP_CLI if that would work, it doesn't have to be anything fancy.)

I added a default value to the "currentBlock" attribute to fix an error I noticed if you insert a carousel block and don't pick a block type.

assets/src/sass/utilities/_theme-supports.scss
pamprn09: @goldenapples those changes to fix color blocks that were not working.

pamprn09:
@goldenapples can we merge this one?

goldenapples:
@pamprn09 yup, it looks good to me!

PR #34, opened 2023-03-07
[#767]: Add "hide page title" control

goldenapples:
Adds a toggle control on the page status box, which toggles whether the page title will be displayed or visually hidden.

image

### DIFF
--- a/assets/src/sass/objects/_header.scss
+++ b/assets/src/sass/objects/_header.scss
@@ -212,7 +212,7 @@ $header-bp-mw: 1049px;
 		margin-top: rem(30);
 		width: 100%;
 
-		body.hide-page-title & {
+		body.hide-page-title & h1 {
 			@extend .visually-hidden
 		}
 	}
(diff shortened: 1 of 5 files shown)


assets/src/sass/objects/_header.scss
```diff
@@ -211,6 +211,10 @@ $header-bp-mw: 1049px;
 		z-index: 2;
 		margin-top: rem(30);
 		width: 100%;
+
+		body.hide-page-title & {
```
MiguelAxcar, line 215: Hiding the `h1` inside `header-content` containing the page title only may be a better option, as it's currently hiding also an `<h2 class="h4 eyebrow">` which has a back link on some pages.

```suggestion
		body.hide-page-title & h1 {
```

I tested the suggested solution here and it's looking good.

image

MiguelAxcar (commented):
That's a very clean and elegant solution, great work @goldenapples ! I dropped a suggestion for your assessment.

goldenapples:
I've made the suggested styling change - nice catch! - and also fixed the meta field permissions so that people who aren't super-admins can edit it. Would you re-review please @MiguelAxcar?

MiguelAxcar (approved):
Tested locally and it's looking great to me!

PR #36, opened 2023-03-08
[#765] Enable dropdowns on desktop for all sections

goldenapples:
Updates the markup so that dropdown menus are initialized properly on desktop, just like on mobile. Ready for code review, but I also want to get some user testing done, since this design isn't at all obvious in the context of drop downs.

These dropdowns work on click or keyboard, but not on hover, and actually as it stands the designs don't really allow for hover interaction without substantial rework.

### DIFF
--- a/assets/src/scripts/modules/dropdown.js
+++ b/assets/src/scripts/modules/dropdown.js
@@ -461,6 +461,13 @@ dropdownMenuItems.forEach( menuItem => {
 } );
 
 /**
+ * Adds class if there's dropdown menu.
+ */
+if ( dropdownMenuItems.lenght !== 0 ) {
+	document.querySelector( 'header' ).classList.add( 'header-default__dropdown-spacing' );
+}
+
+/**
  * Show dropdown.
  *
  * @param {HTMLBodyElement} event document.menu-item['data-dropdown'].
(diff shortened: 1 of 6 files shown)


MiguelAxcar (approved):
Very good work!

Despite needing some final styling, the functionality is working as expected, and the navigation through keyboard follows a good logic:
- if all dropdowns are closed: parent items + carets
- when user open any dropdown: parent item, caret, submenu items
If user closes the submenu, it restores the navigation.

assets/src/scripts/modules/dropdown.js
```diff
@@ -451,6 +451,46 @@ function teardown() {
...
+const dropdownMenuItems = [ ...document.querySelectorAll( '[data-dropdown].menu-item' ) ];
+dropdownMenuItems.forEach( menuItem => {
+	menuItem.addEventListener( 'mouseenter', showDropDown );
+	menuItem.addEventListener( 'mouseleave', mouseLeaveDelay );
+} );
+
+/**
+ * Adds class if there's dropdown menu.
+ */
+if ( dropdownMenuItems.lenght !== 0 ) {
```
goldenapples, line 466:
```suggestion
if ( dropdownMenuItems.length !== 0 ) {
```

varnent:
Looks like this is on hold - correct?

kadamwhite:
Provisionally closing as outdated, and likely to be obviated by the eventual block refresh work

PR #37, opened 2023-03-09
[#763] Improvements for search results page

MiguelAxcar:
This pull request intends to improve search results page.

## Related ticket
https://github.com/humanmade/Wikimedia/issues/763

## Tasks
- [x] Fix bug on block render to make search results to show up
- [x] Set order parameter to 1) relevancy 2) date descending
- [x] Add controls for users to filter search results (all / news (posts) / pages)
- [x] Make search controls accessible
- [x] Added line separator between each result
- [x] Fix bug not displaying H1 title "_Search results for <search-query>_" when querying for some specific words
- [x] Add message "_Showing x - y of z results_"

### DIFF
--- a/assets/src/sass/content/_search-results.scss
+++ b/assets/src/sass/content/_search-results.scss
@@ -3,7 +3,6 @@
 	&__tabs {
 		border-bottom: 1px solid $wmui-color-base70;
 		margin-bottom: 2rem !important;
-		padding-bottom: 6px !important;
 		padding-top: 2rem !important;
 
 		a {
@@ -14,6 +13,7 @@
 			background-color: $wmui-color-base80;
 			text-decoration: none;
 			color: #000;
+			display: inline-block;
 
 			&:hover {
 				background-color: $wmui-color-base90;
--- a/assets/src/sass/content/_search-results.scss
+++ b/assets/src/sass/content/_search-results.scss
@@ -1,0 +1,33 @@
+// Group all style customizations for the 2019-2 transparency report together.
+.search-results {
+	&__tabs {
+		border-bottom: 1px solid #ccc;
+		margin-bottom: 2rem !important;
+		padding-bottom: 0.35rem !important;
+		padding-top: 2rem !important;
+
+		a {
+			padding: 0.5rem 1rem;
+			margin-right: 0.35rem;
+			border: 1px solid #ccc;
+			border-bottom: none;
+			background-color: #eee;
+			text-decoration: none;
+			color: #000;
+
+			&:hover {
+				background-color: #ddd;
+			}
+
+			&.active {
+				background-color: #fff;
+				border-bottom: 1px solid #fff;
+				font-weight: 800;
+			}
+		}
+	}
+}
+
+#search-results .blog-post:not(:last-child) {
+	border-bottom: 1px solid $wmui-color-base80;
+}
--- a/assets/src/sass/content/_search-results.scss
+++ b/assets/src/sass/content/_search-results.scss
@@ -1,7 +1,7 @@
 // Group all style customizations for the 2019-2 transparency report together.
 .search-results {
 	&__tabs {
-		border-bottom: 1px solid #ccc;
+		border-bottom: 1px solid $wmui-color-base70;
 		margin-bottom: 2rem !important;
 		padding-bottom: 6px !important;
 		padding-top: 2rem !important;
@@ -9,19 +9,19 @@
 		a {
 			padding: 0.5rem 1rem;
 			margin-right: 0.35rem;
-			border: 1px solid #ccc;
+			border: 1px solid $wmui-color-base70;
 			border-bottom: none;
-			background-color: #eee;
+			background-color: $wmui-color-base80;
 			text-decoration: none;
 			color: #000;
 
 			&:hover {
-				background-color: #ddd;
+				background-color: $wmui-color-base90;
 			}
 
 			&.active {
-				background-color: #fff;
-				border-bottom: 1px solid #fff;
+				background-color: $wmui-color-base100;
+				border-bottom: 1px solid $wmui-color-base100;
 				font-weight: 800;
 			}
 		}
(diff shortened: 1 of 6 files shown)


assets/src/sass/content/_search-results.scss
```diff
@@ -0,0 +1,33 @@
+// Group all style customizations for the 2019-2 transparency report together.
+.search-results {
+	&__tabs {
+		border-bottom: 1px solid #ccc;
+		margin-bottom: 2rem !important;
+		padding-bottom: 0.35rem !important;
+		padding-top: 2rem !important;
```
goldenapples, line 7: Non-blocking: there's a `rem()` mixin available in the theme which can make it easier to see what these dimensions actually are in pixel units.

assets/src/sass/content/_search-results.scss
```diff
@@ -0,0 +1,33 @@
...
+
+		a {
+			padding: 0.5rem 1rem;
+			margin-right: 0.35rem;
+			border: 1px solid #ccc;
```
goldenapples, line 12: Could we use [theme variables](https://github.com/wikimedia/shiro-wordpress-theme/blob/main/assets/src/sass/helpers/_variables.scss) like `$color-ltgray` or `$wmui-color-base70` for the colors in these styles?

inc/editor/blocks/blog-post.php
```diff
@@ -50,7 +50,11 @@ function render_block( $attributes ) {
 	}
 
 	$post_query = new \WP_Query( [
-		'post_type' => 'post',
+		'post_type' => [ 'post', 'page' ],
+		'orderby'   => [
+			'relevance' => 'DESC',
```
goldenapples, line 55: Does `orderby=relevance` do anything when not performing a search query?

Actually, I don't think any of these parameters are going to be used, since there's a single post ID specified below. Why is this forcing a database query and not just doing a `get_post( $id )`?

search.php
```diff
@@ -6,49 +6,110 @@
...
-
-<?php
+get_header();
 
 $wmf_results_copy = get_theme_mod( 'wmf_search_results_copy', /* translators: 1. search query */ __( 'Search results for %s', 'shiro-admin' ) );
 
 $template_args = array(
 	/* translators: Query that is currently being searched */
-	'h1_title' => sprintf( __( $wmf_results_copy, 'shiro' ), get_search_query() ),
+	'h1_title' => sprintf( __( $wmf_results_copy, 'shiro' ), get_search_query() ), // phpcs:ignore WordPress.WP.I18n.NonSingularStringLiteralText
```
goldenapples, line 15: Yeah, I guess this string shouldn't be getting re-translated, since the default string is already translated above.

search.php
```diff
@@ -6,49 +6,110 @@
...
+				esc_html( $total_results )
+			);
+		}
+		?>
+	</div>
+
+	<div class="search-results__tabs mw-980">
+		<?php
+			$options = [
+				'all' => 'All',
```
goldenapples, line 47: These labels would be nice to mark for translation, so they can be more useful on non-English sites. Could the labels be taken from the post type labels, since they're already internationalized there?

search.php
```diff
@@ -6,49 +6,110 @@
...
+			$option = array_key_exists( $query_option, $options ) ? $query_option : 'all';
+			$selected = esc_attr( $option );
+
+			foreach ( $options as $key => $value ) {
+				$active = $selected === $key ? 'active' : '';
+
+				$href = esc_url( home_url( '/' ) ) . '?s=' . get_search_query();
+				// Simplest way to get the all types is removing post_type param.
+				if ( $key !== 'all' ) {
+					$href .= '&post_type[]=' . $key;
```
goldenapples, line 65: Generally I'd prefer using [add_query_arg](https://developer.wordpress.org/reference/functions/add_query_arg/) to build URLs like these. It's not strictly necessary here since you know all the values this could be adding, but would it make the code here any simpler?

inc/search.php
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Remove post_type = 'profile' from search results.
+ *
+ * @param object $query Full WP_Query object.
+ *
+ * @return void
+ */
+function wmf_remove_profile_from_search( $query ) {
+	if ( $query->is_search() && $query->is_main_query() && ! isset( $_GET['post_type'] ) ) { //phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$query->set( 'post_type', [
```
goldenapples, line 17: This makes sense to do, but I wonder if it'd be simpler to just set `exclude_form_search` in the arguments when registering the profile post type?

https://developer.wordpress.org/reference/functions/register_post_type/#exclude_from_search

goldenapples (commented):
I'll build the styles and give this a full review, but here's my notes from looking over the code.

One suggestion: it looks like this code is skipping any profiles returned in search results, but they're still available to be searched. Do they need to be searched anywhere else? If not, I'd just remove search support from the post type altogether.

assets/src/sass/content/_search-results.scss
MiguelAxcar: Definitely!

inc/editor/blocks/blog-post.php
MiguelAxcar: Your comment made me implement it differently, thanks for the heads up.

MiguelAxcar:
Thank you for your thoughtful review and great advises Than. May you check it again?

inc/search.php
MiguelAxcar: I found appropriate restricting search to posts and pages as client expect, removing stories, profiles and other CPTs that could be added in the future - to appear in the search it needs to be intentionally added.

search.php
MiguelAxcar: Nice tip!

search.php
MiguelAxcar: Sure, added.

goldenapples (approved):
Looks good!

goldenapples (approved):

PR #38, opened 2023-03-10
[#767] Prevent error accessing unregistered meta value

goldenapples:
Fixes an error which shows up in the post-new.php screen for any post types other than pages.

The "wmf_hide_title" value issn't registered for any post types other than Pages, so trying to access it with the `useMeta` hook was throwing errors for any users who weren't super-admins. This moves the hook inside a component which is conditionally rendered, so that it isn't being accessed where it shouldn't be.

### DIFF
--- a/assets/src/editor/plugins/hide-page-title-controls/index.js
+++ b/assets/src/editor/plugins/hide-page-title-controls/index.js
@@ -6,8 +6,20 @@ import { __ } from '@wordpress/i18n';
 import useMeta from '../../hooks/useMeta';
 
 /**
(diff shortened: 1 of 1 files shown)


MiguelAxcar (commented):
@goldenapples I'm still seeing this error when I try to add or edit a post: `The "hide-page-title-control" plugin has encountered an error and cannot be rendered.`

MiguelAxcar (approved):
Working as expected to me! Tested for adding/editing posts, adding/editing pages.

PR #39, opened 2023-03-10
Update composer.json to add support for case-insensitive systems

MiguelAxcar:
Add support for case-insensitive systems such as GNU/Linux.


kadamwhite (approved):

goldenapples (approved):
Nice, I was wondering why I couldn't use those scripts locally, but I hadn't dug into the problem.

PR #41, opened 2023-04-18
#799: Fix Banner Block broken image on Safari

MiguelAxcar:
## Description
This PR intends to fix a bug hiding the Banner Block image on Safari.

## Related ticket
https://github.com/humanmade/Wikimedia/issues/799

## Solution found
I found that the empty `content` property has no effect when applied for an image - it's projected to be used among `::before` and `::after` pseudo-elements. Adding a `!important` as there is a chance of the mixin include having precedence over the explicit declared property.

### Image added on Block Editor
image

### Image missing on Safari
image

## Testing instructions

Edit and Preview this page using Safari:
https://wikimediafoundation-org-develop.go-vip.co/wp-admin/post.php?post=71245&action=edit

The image added should not be missing.

### DIFF
--- a/assets/src/sass/blocks/_banner.scss
+++ b/assets/src/sass/blocks/_banner.scss
@@ -33,6 +33,8 @@
 
 	&__image {
 		@include absolute-full-cover();
+		content: unset !important; // Overwritting content property from the mixin - see #799
+
 		object-fit: cover;
 	}
 


assets/src/sass/blocks/_banner.scss
```diff
@@ -33,6 +33,8 @@
 
 	&__image {
 		@include absolute-full-cover();
+		content: unset !important; // Overwritting content property from the mixin - see #799
```
goldenapples, line 36: Since this is declared after the mixin, it should override the content property from the mixin without this !important.

Its not an issue one way or the other though...

goldenapples (approved):
Nice, good find!

assets/src/sass/blocks/_banner.scss
MiguelAxcar: @goldenapples I thought the same, but I decided to add it just in case. Do you think that it´s safe to remove it?

PR #42, opened 2023-04-20
#797: Fix for content language taxonomy filter

MiguelAxcar:
This PR intends to add a filter to ensure that  posts having different `current-language` taxonomy are not displayed on the post listings.

## Related ticket 
https://github.com/humanmade/Wikimedia/issues/797

## Context
In order to allow creating content in languages that don't have a dedicated site, we register "Content Language" as a taxonomy which can be set on a per-post basis. This allows for individual posts relevant to communities not otherwise served by the blog to be translated and associated with the original post in the same way that MultiLingual Press handles associating content between different language sites.

## Bug description
If a translation post is created and the taxonomy `content-language` is set to a language different than the main language term, currently set to `en_US`.

image

## Solution found
I decided to add a filter to remove from the queries posts which has `content-language` taxonomy set different than the main language term. That is making the undesired posts to disappear from the post listings. Posts which has `en_US` and a different `content-language` taxonomy, such as `pt_BR`, which is incorrect, also disappears from the listings.

### DIFF
--- a/inc/taxonomy-content-language.php
+++ b/inc/taxonomy-content-language.php
@@ -145,8 +145,42 @@ function wmf_add_default_content_language( int $post_ID ): void {
  */
 function wmf_filter_posts_by_content_language( WP_Query $query ) : void {
 
-	// Don't filter the admin, singular pages or non-main queries.
-	if ( is_admin() || ! $query->is_main_query() || is_singular() ) {
+	// Don't filter the admin.
+	if ( is_admin() ) {
+		return;
+	}
+
+	// Restrict content-language filtering to the main query.
+	if ( ! $query->is_main_query() ) {
+		return;
+	}
+
+	// Only filter posts.
+	if ( ! empty( $query->get( 'post_type' ) ) && ! in_array( $query->get( 'post_type' ), apply_filters( 'wmf_content_language_post_types', [ 'post' ] ), true ) ) {
+		return;
+	}
+
+	// Allow user to see single posts no matter what.
+	if ( is_singular() ) {
+		return;
+	}
+
+	// Allow users to see post history setting the content-language.
+	if ( $query->is_tax( 'content-language' ) ) {
+		return;
+	}
+
+	// Try to cover all the potential cases where we don't want the content-language filter to work.
+	if (
+			$query->is_post_type_archive() ||
+			$query->is_category() ||
+			$query->is_tag() ||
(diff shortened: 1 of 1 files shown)


inc/taxonomy-content-language.php
```diff
@@ -134,6 +134,56 @@ function wmf_add_default_content_language( int $post_ID ): void {
...
+ * This is used to ensure that posts having different current-language
+ * taxonomy are not displayed on the post listings.
+ *
+ * @param \WP_Query $query Query object to filter.
+ * @return void
+ */
+function wmf_filter_posts_by_content_language( WP_Query $query ) : void {
+
+	// Don't filter the admin, singular pages or non-main queries.
+	if ( is_admin() || ! $query->is_main_query() || is_singular() ) {
```
goldenapples, line 149: Since the content language taxonomy is only registered for posts, you may be able to also return early if the query is for other post types.

goldenapples (changes requested):
I think there's a few cases you'll need to account for here, in order to allow users to look at language archives like https://wikimediafoundation.org/?content-language=pt.

I also left some questions about performance - it's not terrible as is, but I think avoiding NOT IN queries would be safer and avoid overloading the database... especially since this will be run on basically every archive query on the site.

inc/taxonomy-content-language.php
MiguelAxcar: Excellent suggestions.

goldenapples (approved):
:+1: Looks good!

inc/taxonomy-content-language.php
```diff
@@ -134,6 +134,56 @@ function wmf_add_default_content_language( int $post_ID ): void {
...
+	if ( ! is_array( $tax_query ) ) {
+		$tax_query = [];
+	}
+
+	// Remove from query posts which are related to language terms other than main_locale.
+	$tax_query[] = [
+		'taxonomy' => 'content-language',
+		'field'    => 'slug',
+		'terms'    => wp_list_pluck( $other_language_terms, 'slug' ),
+		'operator' => 'NOT IN',
```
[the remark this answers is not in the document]
MiguelAxcar: That was my first approach, but some side effects were being generated as my code was missing important filtering restrictions. After your suggestions I had a detailed look for applicable restrictions and it's working nicely now. Thank you @goldenapples !

PR #51, opened 2023-04-26
779: Permit assets to be loaded from first available manifest

kadamwhite:
This allows assets to be drawn from different manifests depending on what is available. That way it is possible to use the live-reloading dev server while developing blocks alongside previously-built static theme assets, or vice versa, where previously there was no way to load some prod assets while the dev server was running.

This makes the `npm run start:webpack:theme` and `npm run start:webpack:editor` aliases work as expected.


goldenapples (approved):
This works as I'd expect, thanks for fixing this! Only note is the missing param documentation that phpcs is complaining about.

kadamwhite:
@goldenapples addressed that issue

goldenapples (approved):

PR #52, opened 2023-05-04
#803: Applying fixes for Landing Page Hero CTA button

MiguelAxcar:
This PR intends to fix styling issues on Landing Page Hero block CTA button.

## Related ticket
https://github.com/humanmade/Wikimedia/issues/803

## Changes made
- Automatically applying default styling for CTA button considering block style (Light, Gray, Dark, etc) - checked all of them and it's all working as expected
- Renamed the Editor component to "_Custom CTA styling_"
- Made Editor component "_Custom CTA styling_" to overwrite default styling for CTA button
- Added a default option "_Theme default_" for the CTA styling component on the Editor, which don't overwrite the default CTA button styling
- Tracked and fixed the styling overlapping issues
- Reduced the padding of the info icon SVG, it was impossible to see it inside buttons

## Video demo
https://www.loom.com/share/547a173682474421a50168cdb1044d41

## Testing page on `develop`

To test, just edit this page, try different buttons and test the modifications.

### Admin edit page
https://wikimediafoundation-org-develop.go-vip.co/wp-admin/post.php?post=71309&action=edit

### Frontend page
https://wikimediafoundation-org-develop.go-vip.co/test-fix-803-landing-page-hero-cta-button-styling/

Considering that there are a whole set of buttons available, I don't see the need of adding more options to the Custom CTA styling" Editor component.

Do you @varnent ? Please let me know and I add it.


assets/src/editor/blocks/landing-page-hero/index.js
```diff
@@ -86,12 +86,12 @@ export const settings = {
 		buttonText: {
 			type: 'string',
 			source: 'html',
-			selector: '.hero__call-to-action',
+			selector: '.hero__cta-button',
```
goldenapples, line 89: The selector `.hero__cta-button` isn't in the markup output from the save() method... should this value be changing? It looks like it's still `.hero__call-to-action`.

assets/src/editor/blocks/landing-page-hero/index.js
```diff
@@ -86,12 +86,12 @@ export const settings = {
...
 		},
 		buttonLink: {
 			type: 'string',
 			source: 'attribute',
-			selector: '.hero__call-to-action',
+			selector: 'two .hero__cta-button',
```
goldenapples, line 94: Same here... and also, what is this `two` doing here? Maybe testing code left in accidentally...?

assets/src/sass/blocks/_landing-page-hero.scss
```diff
@@ -119,55 +119,52 @@
...
+			color: #000 !important;
+			background: unset !important;
 			background-color: transparent;
-			border: 2px solid white;
-			background-image: url("../../../../assets/src/svg/individual/info.svg");
+			border: 2px solid black !important;
+
+			&::before {
+				@include button-icon-rules;
+				background-image: url("./assets/src/svg/individual/info.svg") !important;
```
goldenapples, line 146: hmm.... what's different about the way these URLs are referenced? Was the path just wrong before or did something change in the build process? I'm wondering if you had to make this change here, if we need to make the same update everywhere else that we reference icon files by relative paths in sass...

assets/src/svg/individual/info.svg
```diff
@@ -1 +1 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path d="M12 3.2c-4.8 0-8.8 3.9-8.8 8.8 0 4.8 3.9 8.8 8.8 8.8 4.8 0 8.8-3.9 8.8-8.8 0-4.8-4-8.8-8.8-8.8zm0 16c-4 0-7.2-3.3-7.2-7.2C4.8 8 8 4.8 12 4.8s7.2 3.3 7.2 7.2c0 4-3.2 7.2-7.2 7.2zM11 17h2v-6h-2v6zm0-8h2V7h-2v2z" /></svg>
+<svg version="1.1" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><path d="m12 0.19202c-6.4084 0-11.749 5.2331-11.749 11.808 0 6.4407 5.2069 11.808 11.749 11.808 6.4084 0 11.749-5.2331 11.749-11.808 0-6.4407-5.3404-11.808-11.749-11.808zm0 21.469c-5.3404 0-9.6127-4.428-9.6127-9.6611 0-5.3673 4.2723-9.6611 9.6127-9.6611 5.3404 0 9.6127 4.428 9.6127 9.6611 0 5.3673-4.2723 9.6611-9.6127 9.6611zm-1.3351-2.952h2.6702v-8.0509h-2.6702zm0-10.735h2.6702v-2.6836h-2.6702z" stroke-width="1.3384"/></svg>
```
goldenapples, line 1: Is the icon the same? The file looks a lot different, I'm wondering what the change was.

goldenapples (changes requested):
Left a few notes and questions. The issue where the attribute sources are broken for button text and URL should be fixed, because otherwise those fields just get lost whenever the block is re-edited. Otherwise I'm just wondering about the icons.

I like the "theme default" option being the default, rather than some hardcoded styles that don't match darker block styles.

I'm noticing we probably need a better way to include icons in these buttons so they can be styled - including the SVG as a background image doesn't work in some color combinations with a dark background, but if it were inlined as an SVG it could be styled in the same color as the button text. Is there already some functionality for including icons in buttons somewhere in shiro? I can't remember if someone already built that out - if not, it's not necessary to build it out for this one ticket, but if it exists, there might be a way to use it.

assets/src/editor/blocks/landing-page-hero/index.js
MiguelAxcar: Definitely Than, both were left as a mistake. Thanks for catching this!

assets/src/sass/blocks/_landing-page-hero.scss
MiguelAxcar: I found other places referencing as this. Such as https://github.com/wikimedia/shiro-wordpress-theme/blob/main/assets/src/sass/blocks/_button.scss#L16
I believe that it was an error.

assets/src/svg/individual/info.svg
MiguelAxcar: I slightly reduced the padding of the image, it was too small inside buttons. 

## New
image

## Old
image

MiguelAxcar:
@goldenapples thank you for the code review. Currently the SVG icon has the same color of the text, but I agree that we can find a better way for doing it, allowing styling the SVG itself. May we open another ticket? A broader suggestion for a future ticket would to improve the CTA component, allowing customization of background color, border size/color, icon, size, etc with few pre-filled options that would work changing the selectboxes only.

MiguelAxcar:
`develop` was updated with the last changes.

goldenapples:
> May we open another ticket? A broader suggestion for a future ticket would to improve the CTA component, allowing customization of background color, border size/color, icon, size, etc with few pre-filled options that would work changing the select boxes only.

Yeah, those enhancements can definitely be addressed in a separate ticket. I didn't mean to hold up this fix trying to improve something unrelated.

We do have an [IconSelector component](https://github.com/wikimedia/shiro-wordpress-theme/blob/803-landing-page-hero-block-fix-cta-button-styles/assets/src/editor/blocks/core-button/IconSelector.js) which I was thinking could be used here instead of a separate hard-coded list of combinations of icon and color options. But it looks like that control is also just using the svg as a background image, so it wouldn't address the issue I was hoping to fix.

PR #55, opened 2023-05-11
#768: Improve filtering on news section

MiguelAxcar:
This PR intends to add a filter component for News section, allowing user to filter post list by text, date interval and categories.

The ticket asks for filtering by type (Press release, story, interview, etc.) and country. 

For type, I refer to this comment: https://github.com/humanmade/Wikimedia/issues/763#issuecomment-1462466007

This specific part:

> You asked for giving controls for user to search "all, or pages, or news", but **1)** there is no way for setting a post as "news" **2)** the "post" wasn't included and **3)** there is a custom post type on Wikimedia Foundation site named "story" - see [stories on frontend](https://wikimediafoundation.org/about/transparency/2022-1/stories-2022-1/#section-1) and see [stories on backend](https://wikimediafoundation.org/wp-admin/edit.php?post_type=story) (link "Stories" in the admin left menu). So I found more appropriate changing it to: "All, Articles, Pages and Stories", referring to "posts" as "articles" - is it ok?

I see no way of using post types, and I believe that it overlaps with categories, as I found categories for stories, interviews and press releases.

For countries, I didn't find a way for filtering by country, that does not make sense for me as there is no way of relating a post to a country. 

My idea was wrapping up the filters that I had, and capture some feedback from PO that can be subject for another ticket.

Also, some work for internationalization using the custom translation system for WMF is needed, and I'm not sure about adding it to this ticket.

## Testing instructions
- Pull the code and access /news through the menu

## Closed filters - Desktop view
image

## Opened filters - Desktop view
image

## One filter applied - Desktop view
image
image
image

## One filter applied - Mobile view
image

## Filters opened - Mobile view
image


kadamwhite (changes requested):
Nice UI, Miguel! Can you tag @varnent on the ticket https://github.com/humanmade/wikimedia/issues/768 so he knows to review your loom video and comment on the UX? Or could tag him in Slack.

I've noted a few areas that could use some simplification and cleanup, and unfortunately I need to ask for you to overhaul the CSS to be less over-specified. I've also tagged Than with a question about whether I'm blanking on a WP function which might help us with the HTML output on this form.

template-parts/post-list-filters.php
```diff
@@ -0,0 +1,143 @@
...
+
+			<div class="filter-by-date">
+				<h5>
+					<?php printf( esc_html__( 'Filter by date', 'shiro' ) ); ?>
+					<?php if ( ! empty( $_GET['date_from'] ) && ! empty( $_GET['date_to'] ) ) : ?>
+						&nbsp;<i>(<?php printf( esc_html__( 'applied', 'shiro' ) ); ?>)</i>
+					<?php endif; ?>
+				</h5>
+				<div class="filter-date-inputs-container">
+					<input type="date" name="date_from" placeholder="From" value="<?php echo isset( $_GET['date_from'] ) ? esc_attr( sanitize_text_field( $_GET['date_from'] ) ) : ''; ?>">
```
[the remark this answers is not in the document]
goldenapples: Yeah, unless we register "date_from" as a query_var, `get_query_var()` won't work here. I'm not aware of a built-in function for retrieving arbitrary GET parameters.

My personal preference is to read all of the query args you'll need into variables at the top of the template part, so that it's obvious what global state it depends on. So at the top of the file you might have something like

```
$date_from = $_GET['date_from'] ?? '';
$date_to = $_GET['date_to] ?? '';
$categories = $_GET['categories'] ?? [];
```

...and then escape them on output.

(This is personal preference though, and I don't think there's anything wrong with the way you've done it or abstracting to the utility function like @kadamwhite suggested.) 

One thing to note though is to be aware of the difference between sanitization and escaping. Use sanitization if you're saving a value to the database, and escaping if you're outputting in HTML. I don't think this `sanitize_text_field` call is necessary here. What might be necessary is `wp_unslash()`, though, to strip slashes that WP adds to all of the REQUEST fields.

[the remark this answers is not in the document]
MiguelAxcar: Thanks @goldenapples, that leaded me to few simplifications more, and I absorbed your pattern of parsing the query vars on the template part header :+1:

assets/src/sass/components/_post-list-filters.scss
```diff
@@ -0,0 +1,164 @@
...
+		margin-left: auto;
+
+		@media (max-width: $small-bp) {
+			margin: 1rem;
+		}
+
+		&::before {
+			background-image: url("./assets/src/svg/individual/filter.svg");
+			background-position: center;
+			background-repeat: no-repeat;
```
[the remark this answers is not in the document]
MiguelAxcar: Nice!

I copied that from `@mixin button-icon-rules` available on 
https://github.com/wikimedia/shiro-wordpress-theme/blob/main/assets/src/sass/blocks/_button.scss#L15-L27

Ended up moving it to theme button mixins and reusing it, applying the cleanup there, as you'll se on the code.

Btw, despite MDN, `contain` does not work at the moment.

assets/src/sass/components/_post-list-filters.scss
```diff
@@ -0,0 +1,164 @@
...
+				font-family: $font-family-system-sans;
+				font-size: 0.75rem;
+				background-color: $color-periwinkle;
+				border-radius: 0.25rem;
+				color: $color-blue;
+				font-weight: 700;
+				margin-bottom: 0.5rem;
+				margin-left: 0.5rem;
+				padding: 0.25rem;
+				padding-right: 0.75rem;
```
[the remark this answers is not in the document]
MiguelAxcar: That was intentional, as in my opinion it was clearer to read like I did. Suggestion accepted.

assets/src/scripts/post-list-filters.js
```diff
@@ -0,0 +1,83 @@
...
+ * @param {HTMLElement} filterButton - The button used to toggle the filter container.
+ * @param {HTMLElement} filterContainer - The filter container.
+ */
+const toggleFilterContainer = ( filterButton, filterContainer ) => {
+	/**
+	 * Click event handler for the filter button.
+	 */
+	const clickHandler = () => {
+	  const isHidden = filterContainer.style.display === 'none' || filterContainer.style.display === '';
+	  filterContainer.style.display = isHidden ? 'flex' : 'none';
```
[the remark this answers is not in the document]
MiguelAxcar: Definitely, great idea. Applied.

assets/src/scripts/post-list-filters.js
```diff
@@ -0,0 +1,83 @@
...
+
+	  const buttonLabel = filterButton.getAttribute( 'show-filters-applied-button-label' );
+	  filterButton.innerHTML = isHidden ? 'Hide filters' : buttonLabel;
```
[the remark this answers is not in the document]
MiguelAxcar: Awesome, much better.

assets/src/scripts/post-list-filters.js
```diff
@@ -0,0 +1,83 @@
...
+};
+
+/**
+ * Initializes the post list filters functionality.
+ */
+const initializePostListFilters = () => {
+
+	// Controls filters container visibility toggle.
+	const filterButton = document.querySelector( '.post-list-filter__toggle' );
+	filterButton.setAttribute( 'show-filters-applied-button-label', filterButton.innerHTML );
```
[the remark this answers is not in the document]
MiguelAxcar: Great, thanks!

assets/src/scripts/post-list-filters.js
```diff
@@ -0,0 +1,83 @@
...
+	const filterContainer = document.querySelector( '.post-list-filter__container' );
+	toggleFilterContainer( filterButton, filterContainer );
+
+	// Controls date filters reset.
+	const resetDateFiltersButton = document.querySelector( '.button-reset-date-filters' );
+	const fromDateInput = document.querySelector( 'input[name="date_from"]' );
```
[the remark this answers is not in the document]
MiguelAxcar: Good suggestion, thanks - applied to the only possible element.

inc/post-list-filters.php
```diff
@@ -0,0 +1,64 @@
...
+	if ( ! $query->is_main_query() ) {
+		return;
+	}
+
+	if ( ! isset( $_GET['post_list_filters_nonce'] ) || ! wp_verify_nonce( sanitize_text_field( $_GET['post_list_filters_nonce'] ), 'post_list_filters' ) ) {
+		return;
+	}
+
+	// Filter by search text.
+	$search_text = isset( $_GET['s'] ) ? sanitize_text_field( $_GET['s'] ) : '';
```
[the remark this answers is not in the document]
MiguelAxcar: You are correct, adding the search string to the query `$_GET['s']` is set is a built-in WordPress function . Thanks for the heads up.

inc/post-list-filters.php
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Handles post list filtering by text, date and categories.
+ *
+ * This function is hooked into the pre_get_posts action below.
+ *
+ * @param WP_Query $query The WP_Query instance.
+ * @return void
+ */
+function custom_post_list_filters( $query ) : void {
+	if ( ! $query->is_main_query() ) {
```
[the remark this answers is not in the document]
MiguelAxcar: Nice, added few guards more.

template-parts/post-list-filters.php
```diff
@@ -0,0 +1,143 @@
...
+
+		<h3>
+			<?php
+			$total_results = $wp_query->found_posts;
+			$paged = ( get_query_var( 'paged' ) ) ? get_query_var( 'paged' ) : 1;
+			$posts_per_page = get_query_var( 'posts_per_page' );
+			$first_result = ( $posts_per_page * $paged ) - $posts_per_page + 1;
+			$last_result = min( $total_results, $wp_query->post_count * $paged );
+			if ( $total_results === 1 ) {
+				printf( esc_html__( 'Showing 1 of 1 result', 'shiro' ) );
```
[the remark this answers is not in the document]
MiguelAxcar: I don't see it being applied here, as the message is for 
- No results, no message
- single result
- more than one result

But I will be available for any opposite opinions.

template-parts/post-list-filters.php
```diff
@@ -0,0 +1,143 @@
...
+					$filter_count += count( $_GET['categories'] );
+				}
+
+				printf( esc_html__( 'Show filters', 'shiro' ) );
+
+				if ( $filter_count > 0 ) {
+					echo ' <i>';
+					/* translators: 1. how many filters were applied */
+					printf( esc_html__( '(%s applied)', 'shiro' ), esc_html( $filter_count ) );
+					echo '</i>';
```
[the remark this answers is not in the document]
MiguelAxcar: Great!

In terms of UI, I'm wondering if the `(<em>applied</em>)` if enough. I was wondering if adding an extra `<u>` or `<mark>` wouldn't be better. What do you think?

## Using `<mark>` 
image

## Using `<u>` (maybe can be considered a link by users, IDK)
image

template-parts/post-list-filters.php
```diff
@@ -0,0 +1,143 @@
...
+
+		<?php wp_nonce_field( 'post_list_filters', 'post_list_filters_nonce' ); ?>
+
+		<div class="post-list-filter__container mw-980">
+
+			<div class="filter-by-text">
+				<h5>
+					<?php printf( esc_html__( 'Filter by text', 'shiro' ) ); ?>
+					<?php if ( ! empty( get_search_query() ) ) : ?>
+						&nbsp;<i>(<?php printf( esc_html__( 'applied', 'shiro' ) ); ?>)</i>
```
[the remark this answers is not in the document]
MiguelAxcar: Great, done.

assets/src/sass/blocks/_button.scss
```diff
@@ -12,20 +12,6 @@
...
-	background-size: contain;
-	content: "";
-	display: inline-block;
-	height: 1em;
-	margin-right: 0.5em;
-	position: relative;
-	top: 2px;
-	width: 1em;
-}
-
```
MiguelAxcar, line 28: I moved it from here to `_mixins-buttons.scss` as you'll see below, so it can be reused.

assets/src/sass/components/_post-list-filters.scss
```diff
@@ -0,0 +1,140 @@
+.post-list-filter {
```
MiguelAxcar, line 1: Restructured the file looking for less nesting.

template-parts/post-list-filters.php
```diff
@@ -0,0 +1,149 @@
...
+ * Template for post list filter controls
+ *
+ * @package shiro
+ */
+
+// Get query vars.
+$query_var_search_term = get_search_query();
+$query_var_date_from   = isset( $_GET['date_from'] ) ? sanitize_text_field( $_GET['date_from'] ) : '';
+$query_var_date_to     = isset( $_GET['date_to'] ) ? sanitize_text_field( $_GET['date_to'] ) : '';
+$query_var_categories  = isset( $_GET['categories'] ) ? array_map( 'sanitize_text_field', $_GET['categories'] ) : [];
```
MiguelAxcar, line 12: Thanks @goldenapples , I absorbed your preference as it's a good idea, allow reusing the vars and avoid applying sanitization many times.

MiguelAxcar:
Your thoughtful review was very much appreciated @kadamwhite , thank you. May you please check again to see if everything was addressed as you expected?

assets/src/scripts/post-list-filters.js
```diff
@@ -0,0 +1,82 @@
...
+ */
+const toggleFilterContainer = ( filterButton, filterContainer ) => {
+	/**
+	 * Click event handler for the filter button.
+	 */
+	const clickHandler = () => {
+		const isHidden = ! filterContainer.classList.contains( 'post-list-filter__container--open' );
+		filterContainer.classList.toggle( 'post-list-filter__container--open' );
+
+		filterButton.innerHTML = isHidden ? 'Hide filters' : filterButton.dataset.showFiltersAppliedButtonLabel;
```
goldenapples, line 15: This string still needs some kind of internationalization, so it can be displayed in non-English languages.

assets/src/scripts/post-list-filters.js
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Initializes the post list filters functionality.
+ */
+const initializePostListFilters = () => {
+
+	// Controls filters container visibility toggle.
+	const filterButton = document.querySelector( '.post-list-filter__toggle' );
+	filterButton.dataset.showFiltersAppliedButtonLabel = filterButton.innerHTML;
+	const filterContainer = document.querySelector( '.post-list-filter__container' );
+	toggleFilterContainer( filterButton, filterContainer );
```
goldenapples, line 62: Just as a naming nitpick - this function name sounds like it actually toggles the container. I would have named it something like 'addFilterToggleHandlers'...

template-parts/post-list-filters.php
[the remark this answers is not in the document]
goldenapples: Yeah, it doesn't make a difference logically, it's just a nice function for translation, especially when you're translating into languages that have difference rules for pluralization. In this case, it'd give you something like

```
printf(
    _n(
        'Showing %1$d of %3$d result',
        'Showing %1$d&ndash;%2$d of %3$d results',
        $total_results,
        'shiro'
    ),
    $first_result,
    $last_result,
    $total_results
);

```

Ultimately, like you mentioned on standup, all of this is academic because WIkimedia doesn't use a standard translation system. But if they were, this has the benefit of allowing languages that pluralize differently to be translated naturally. (As one example, in Russian there are 3 different plural cases for nouns - one for numbers ending in 1, one for numbers ending in 2-4, and another for numbers ending in 5-0.)

template-parts/post-list-filters.php
```diff
@@ -0,0 +1,149 @@
...
+
+				<h5>
+					<?php printf( esc_html__( 'Filter by category', 'shiro' ) ); ?>
+					<?php if ( count( $query_var_categories ) > 0 ) : ?>
+						&nbsp;<em><u>(<?php echo count( $query_var_categories ) . ' ' . esc_html__( 'applied', 'shiro' ); ?>)</u></em>
+					<?php endif; ?>
+				</h5>
+
+				<ul class='category-container'>
+					<?php foreach ( $categories_array as $category->slug => $category_display ) : ?>
```
goldenapples, line 130: This foreach statement looks weird, does it do what you'd expect? I'd expect the key to have to be a simple variable like `$category_slug`

template-parts/post-list-filters.php
```diff
@@ -0,0 +1,149 @@
...
+							<input type="checkbox" name="categories[]" value="<?php echo esc_attr( $category->slug ); ?>" <?php checked( in_array( $category->slug, $query_var_categories ) ); ?>>
+							<?php echo esc_html( $category_display ); ?>
+						</label>
+					</li>
+					<?php endforeach; ?>
+				</ul>
+
+			</div>
+
+			<button class='action-button action-button--clear' id="button-clear-filters" type="reset"><?php printf( esc_html__( 'Clear filters', 'shiro' ) ); ?></button>
```
goldenapples, line 142: Nitpick: As there's no format strings in this text, the `printf` isn't necessary. You could just do 
```suggestion
			<button class='action-button action-button--clear' id="button-clear-filters" type="reset"><?php esc_html_e( 'Clear filters', 'shiro' ); ?></button>
```

template-parts/post-list-filters.php
```diff
@@ -0,0 +1,149 @@
...
+			</div>
+
+			<div class="filter-by-category">
+				<?php
+				$categories = get_categories();
+
+				foreach ( $categories as $category ) {
+					$category_display = ( $category->parent == 0 )
+						? $category->name
+						: get_category_parents( $category->parent, false, ' > ', false ) . $category->name;
```
goldenapples, line 115: Do you need a space or a separator between the category parents list and the category name? Just curious, I haven't checked what the output of `get_category_parents()` looks like.

template-parts/post-list-filters.php
```diff
@@ -0,0 +1,149 @@
...
+				</h5>
+				<div class="search-text-input-button">
+					<input type="text" name="s" value="<?php echo esc_attr( $query_var_search_term ); ?>">
+				</div>
+			</div>
+
+			<div class="filter-by-date">
+				<h5>
+					<?php printf( esc_html__( 'Filter by date', 'shiro' ) ); ?>
+					<?php if ( ! empty( $query_var_date_from ) && ! empty( $query_var_date_to ) ) : ?>
```
goldenapples, line 97: Should this be AND or OR? Is it possible to search with a "from" data but not a "to" date?

inc/post-list-filters.php
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	if ( is_search() ) {
+		return;
+	}
+
+	// Filter by date interval.
+	$date_from = isset( $_GET['date_from'] ) ? sanitize_text_field( $_GET['date_from'] ) : '';
+	$date_to   = isset( $_GET['date_to'] ) ? sanitize_text_field( $_GET['date_to'] ) : '';
+	if ( ! empty( $date_from ) && ! empty( $date_to ) ) {
```
goldenapples, line 45: I asked elsewhere about this logic - Do you need to specify both a "from" date and a "to" date? It seems like there might be times when you only need to specify one of those.

assets/src/scripts/post-list-filters.js
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Walk through the form elements and reset them.
+ *
+ * @param {HTMLElement} form - The form element.
+ * @returns {void}
+ */
+const resetFormFields = form => {
+
+	// Reset all inputs on form.
+	const inputs = form.querySelectorAll( 'input[type="text"], input[type="date"]' );
```
goldenapples, line 41: This includes the search query field - is that intentional?

goldenapples (approved):
This is looking good, and I'd approve the code as is. 

There's an open question about how to internationalize this - are you going to add .pot files for each language or build in customizer controls? And I know that's part of a bigger architectural decision you're working on, so it doesn't have to be resolved here.

The only thing that really looks wrong to me is the foreach loop in the category container where you're assigning to `$category->slug` in each iteration of the loop - I guess that works but it looks very strange to me!

template-parts/post-list-filters.php
MiguelAxcar: Understood  `_n`, nice trick, thanks!

About the loop, I got the problem - it was _strangely_ working because I was rewriting the object property defined above. Fixed it!

assets/src/scripts/post-list-filters.js
MiguelAxcar: It's actually toggling the filter container, this element marked as red:

image

template-parts/post-list-filters.php
MiguelAxcar: The default separator it `/` without spaces, and I would like having `>` with spaces to look better for more than one word.

template-parts/post-list-filters.php
MiguelAxcar: I changed it to use `date_from`, `date_to` or both. I built it thinking in creating structured date intervals and didn't even thinking using it separately. Good idea, and it works. Thanks @goldenapples !

assets/src/scripts/post-list-filters.js
MiguelAxcar: Yes, that's the action of button "Clean filters" to guarantee that the form is being properly cleaned up entirely before submitting. Different of the action of the button "Reset" which reset only date filters.

MiguelAxcar:
@goldenapples finished CR, applying all the changes you suggested.

MiguelAxcar:
@kadamwhite may you assess if the requested changes were addressed as you was expecting? Thank you!

template-parts/post-list-filters.php
```diff
@@ -0,0 +1,153 @@
...
+
+			<div class="filter-by-category">
+				<?php
+				$categories = get_categories();
+				foreach ( $categories as $category ) {
+					$category_display = ( $category->parent == 0 )
+						? $category->name
+						: get_category_parents( $category->parent, false, ' > ', false ) . $category->name;
+
+					$categories_array[ $category->slug ] = $category_display;
```
kadamwhite, line 121: So this is interesting -- I didn't know that you could implicitly create an array like this in PHP! What I mean is, that if you do this,
```php
<?php

$arr['something'] = 'foo';
print_r( $arr );
```
you get
```
Array
(
    [something] => foo
)
```
even if `$arr` wasn't previously defined. So, this line is creating `$categories_array` as an array implicitly.

I'd recommend making this implicit behavior more explicit, by declaring `$categories_array = [];` before the `foreach` loop.

I'd also suggest moving this array computation up to the initial PHP block at the top of the template partial, along with the other template variables.

assets/src/scripts/post-list-filters.js
kadamwhite: @MiguelAxcar Please flag this one for Greg in testing, in my own experimentation I found that I didn't expect this field to be cleared when I clicked this. I'd recommend seeing what @varnent feels is best.

kadamwhite (approved):
Nice cleanup, thanks @MiguelAxcar! Only thing left is to [flag this comment](https://github.com/wikimedia/shiro-wordpress-theme/pull/55/files#r1194075347), I felt like it was unexpected to me when testing that the Clear button would also delete the query itself and think we should ask Greg what feels best to him. Please tag him with that question in the main issue, if you could.

MiguelAxcar:
Thank you @kadamwhite - Following your suggestion, I asked special attention to Greg when reviewing this particular part of the issue on my [last comment in the ticket](https://github.com/humanmade/Wikimedia/issues/768#issuecomment-1548484083).

kadamwhite (approved):
Thanks, @MiguelAxcar ! I appreciate the pointer to where the question had been asked.

PR #57, opened 2023-05-15
814: Server-render TOC block for third-level items

kadamwhite:
<details>

<summary>RFC on this approach, originally included in the description here</summary>

I've baked this down to the reduced POC of my idea to get a direction check from @goldenapples, @MiguelAxcar and possibly @joeleenk. The ticket humanmade/wikimedia#814 wants to be able to support third-level items, and while working on the ticket, I began to explore moving the rendering of the block to the server wholesale. This could decouple the block from a strict hierarchy (which Greg has noted is error prone, in a comment on the original ticket, e.g. you can't use the ToC block within a reusable block) — at the expense of codifying some rules such as "there must only be one on a page."

The way this branch works is,

- Build a list of heading blocks while rendering the post's blocks
- (There's an improvement step that could be added here to rebuild the headings list into a better nested structure, to facilitate better iteration and template code)
- Render a nested list of anchor links pointing to all found headings, and replace the rendered output of the block with that list of headings identified while rendering blocks

If we continued down this road we could remove some of the JS save code, but I want to get feedback on the idea before proceeding. It is working in my local and successfully generates the next level item (CSS and JS needed for best functionality), but I wanted to check whether we had any concerns about the overall proposal or whether it is appropriate to move this to server-side rendering. It feels like a lot less work process-wise to do this on the server than on save, but does mean more "work" during rendering.

</details>

## Support third-level headings in single-page Table of Contents block

This PR refactors the existing ToC block to be server-rendered, and adds support for rendering third-level menu items.

**known issues**

The third-level items are not currently represented in the visual editor. They will display on the frontend if the appropriate option is selected.

### DIFF
--- a/assets/src/editor/blocks/table-of-contents/deprecations.js
+++ b/assets/src/editor/blocks/table-of-contents/deprecations.js
@@ -1,0 +1,160 @@
+/**
+ * Handle old versions of this block.
+ *
+ * @see https://developer.wordpress.org/block-editor/reference-guides/block-api/block-deprecation/
+ */
+import { useBlockProps } from '@wordpress/block-editor';
+import { __ } from '@wordpress/i18n';
+
+import HeadingLinks from './HeadingLinks';
+
+/**
+ * Original version of ToC block, which rendered JS-side within the editor
+ * and stored the rendered headings list in post content as well as in block
+ * attributes. Transitioned in May 2023 to v2, where the block is entirely
+ * server-rendered in PHP.
+ */
+const v1 = {
+	supports: {
+		inserter: false,
+		multiple: false,
+		reusable: false,
+	},
+	attributes: {
+		headingBlocks: {
+			type: 'array',
+			default: [],
+		},
+	},
+	/**
+	 * Render save of the original v1 table of contents block.
+	 */
+	save( { attributes } ) {
+		const blockProps = useBlockProps.save( {
+			className: 'table-of-contents toc',
+		} );
+
+		return (
+			<>
+				{ attributes.headingBlocks.length > 0 && (
+					<nav
+						className="toc-nav"
+						data-backdrop="inactive"
+						data-dropdown="toc-nav"
+						data-dropdown-content=".toc"
+						data-dropdown-status="uninitialized"
+						data-dropdown-toggle=".toc__button"
+						data-sticky="false"
+						data-toggleable="yes"
+						data-trap="inactive"
+						data-visible="false"
+					>
+						<h2 className="toc__title screen-reader-text">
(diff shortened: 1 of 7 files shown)


goldenapples:
I don't know how this block was being managed before. I do like the idea of it being server-rendered though; that just makes sense to me. 

I don't know about all the regex and DOMDocument stuff going on here - couldn't the whole thing be done by parsing blocks? That just seems simpler and less error-prone to me. But maybe this is just an in-between state that's trying to add a bit of server-side logic while still keeping the client side controls.

Questions: is the ToC always going to only be parsing for header blocks? and is the ToC entry always going to be the text of the header block? Those assumptions might be safe in some content but not in others. As one alternate suggestion, there's an "HTML Anchor" field in the core block controls for blocks that support the "anchor" attribute. It might makes sense to look for blocks with values in that attribute (maybe in combination with an editor plugin that adds a default value for "anchor" to certain blocks).

kadamwhite:
> I don't know about all the regex and DOMDocument stuff going on here - couldn't the whole thing be done by parsing blocks?

The issue is that `parse_blocks` on the server (or the parsed block object you get in the render or pre_render hook) doesn't include discrete values for anchor ID, content, etc, for HTML content. WordPress stores the `h2` as `<h2 id="whatever">contents</h2>`, and the only time that gets mapped into `{ anchor, content }` is client-side when JS parses the block content itself (core uses DOM selectors to target the values of those attributes). This means that to "safely" get the values server-side, we have to replicate using DOMDocument the parsing that happens client-side in JS.

**Question 1**, I'm not sure whether we should consider DOMDocument a performance issue? (I'm currently planning to cross that bridge once we get there, it seems snappy enough locally.) I could do all this with regex fairly easily since the markup of a heading block is pretty regimented, if we wanted to avoid DOMDocument.

> is the ToC always going to only be parsing for header blocks? and is the ToC entry always going to be the text of the header block?

That's how it works today, so I believe it's a safe assumption for now and enhancing it to handle other anchor'd, non-header blocks could be done in a future enhancement.

> I do like the idea of it being server-rendered though; that just makes sense to me.

Thanks, that was my biggest concern, was that this was the wrong place to be doing this.

kadamwhite:
@MiguelAxcar @varnent this is now ready for review, apologies that it's coming only at the very end of the sprint.

- There is now a toggle in the single-page version of the ToC block to include H3s
- If toggled, they will render on the frontend
- Rendering has moved to PHP

I burned more time than I'd like trying to refactor more fundamental things about the theme to make it easier to do a server-rendered preview of the block, so at the moment you will not see the third-level items _in the editor_. They should show on the frontend.

I ran out of time for the linked version.

varnent (approved):

PR #58, opened 2023-05-16
#825: Fix and improve Customizer page selection interface

MiguelAxcar:
This PR intends to fix and improve the page selection on Customizer.

### Fixed
- Removed the limit of 100 pages, as Wikimedia currently has more than 150 pages. 

### Improved
I found it very hard finding the pages, full of duplication, so I decided to:
- Sort it alphabetically
- Remove pages with empty title from the list
- Display only published pages
- Add page ID (to give user the capacity of looking for page ID if needed)
- Add page relative URL (as many pages has the same title)

image

### DIFF
--- a/inc/classes/customizer/class-base.php
+++ b/inc/classes/customizer/class-base.php
@@ -67,7 +67,7 @@ abstract class Base {
 		$posts = new WP_Query(
 			array(
 				'post_type'      => 'page',
-				'posts_per_page' => -1,
+				'posts_per_page' => 1000, // phpcs:ignore WordPress.WP.PostsPerPage.posts_per_page_posts_per_page
 				'orderby'        => 'title',
 				'order'          => 'ASC',
 				'post_status'    => 'publish',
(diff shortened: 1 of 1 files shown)


inc/classes/customizer/class-base.php
```diff
@@ -67,12 +67,18 @@ public function page_choices() {
...
+				'orderby'        => 'title',
+				'order'          => 'ASC',
+				'post_status'    => 'publish',
 			)
 		);
 
 		foreach ( $posts->posts as $post_choice ) {
-			$choices[ $post_choice->ID ] = $post_choice->post_title;
+			if ( ! empty( $post_choice->post_title ) ) {
+				$page_relative_link = str_replace( home_url(), '', get_permalink( $post_choice->ID ) );
```
goldenapples, line 79: There's a core function [wp_make_link_relative()](https://developer.wordpress.org/reference/functions/wp_make_link_relative/) that might do what you want here. It's not exactly the same, because it doesn't handle sites where the home URL is in a subdirectory, but it's similar.

goldenapples (approved):
Nice. This looks like it would make this post selector a lot easier to use.

MiguelAxcar:
Your suggestions were added @goldenapples , thank you for the review.

PR #69, opened 2023-06-07
#873: Quickfix to disable adding H3 to TOC

MiguelAxcar:
This PR intends to add a production quickfix to disable displaying H3 on TOC block.


ajvillegas (approved):
Look good

PR #75, opened 2023-07-18
#876: Bugfix sorting executives and experts on role/profile page

MiguelAxcar:
This PR intends to fix issue of sorting executives and experts by last name meta field, on role list page.

I invested significant effort in resolving this bug, targeting sorting by `last_name` meta field. Initially, I attempted to utilize the existing sorting feature, but after analyzing the code and considering various scenarios, I found that the shortest path was to implement a custom sorting function specifically for handling executives and experts sorting.

Also, an error was happening because the action was trying to run `phpcs` on files that were deleted in the pull request, then I modified the Lint to run over changed files only using GIT.


inc/template-functions.php
```diff
@@ -130,6 +130,8 @@ function wmf_get_role_posts( $term_id ) {
 			'post_type'      => 'profile',
 			'fields'         => 'ids',
 			'orderby'        => 'title',
+			'meta_key'       => 'last_name',
+			'orderby'        => 'meta_value',
```
MiguelAxcar, line 134: I changed to this, so `inc/classes/profiles/class-sorter.php` seems to be no longer needed. Not related to the bug fixed on the PR, but decided to keep it as sorting  directly instead of sorting manually can potentially increase performance.

.github/workflows/php.yml
```diff
@@ -52,6 +50,11 @@ jobs:
         if: ${{ steps.paths.outputs.phpcs == 'true' }}
         run: parallel-lint --exclude .git --exclude .github --exclude vendor .
 
+      - name: Fetch base and head branches
+        run: |
+          git fetch origin ${{ github.base_ref }}:${{ github.base_ref }}
+          git fetch origin ${{ github.head_ref }}:${{ github.head_ref }}
```
goldenapples, line 56: :thinking: Does this speed up the CI pipeline by only checking out the branches needed?

inc/template-functions.php
goldenapples: You've got two "orderby" parameters in this query; I think the first "orderby => title" can be removed.

goldenapples (approved):
Other than the extra "orderby" parameter, this looks good to go.

It looks like now the posts should be returned from the query in alphabetical order, but then you're resorting them with `strnatcasecmp`. Do we need both of those sorts? 

In the past, WordPress VIP's coding standards discouraged meta queries, because joining tables in the database query was slower than sorting them in PHP. This is mainly an issue with sites that (1) have massive amounts of content in the meta tables, and (2) have a lot of database traffic. I don't know how the different options compare performance-wise in this case, and this page is low-enough traffic that query performance is probably not a big issue. But there might be a way to test out the performance of the SQL orderby as compared to the PHP sort, if you're interested.

Getting rid of the Sorter class is nice - this code is much simpler and easier to read.

MiguelAxcar:
> It looks like now the posts should be returned from the query in alphabetical order, but then you're resorting them with strnatcasecmp. Do we need both of those sorts?

What you said makes total sense @goldenapples, and this was my first try. After studying the code  I found two different relationships between Profiles and Roles.

### 1) When editing the profile
image
This feature is used to relate profiles and roles, and it's used to 1) populate the multi-checkbox component on role edit page and 2) all the profiles are displayed if no executives or experts are selected. I'm not aware other places which uses it. 

### 2) When editing the role
image
This feature is used to highlight, among all profiles related to the role, the ones which needs to be highlighted on the page as executives and experts. The profiles displayed inside the multi-checkbox component are being displayed sorted by last name, but the order or their selection where used as the sorting parameter.

Thanks!

varnent (approved):

.github/workflows/php.yml
MiguelAxcar: I believe that fetching only necessary branches can make CI pipeline to run faster, specially on larger repositories. But it's not just about speed, I think it's more semantic tough.

PR #78, opened 2023-07-20
#885: Add "frame-src" directives to Content-Security-Policy header

goldenapples:
Adds YouTube and Vimeo domains to the allowed frame sources as specified in the content security policy.


varnent (approved):
Looks good - thank you @goldenapples!

PR #81, opened 2023-08-17
#891: Allow scrolling to view all rotating headers in home hero block

goldenapples:
Enables scrolling inside the container which holds the rotating headings in the home-hero block's edit component, so that it's possible to see (and edit) the text if more headings are set than can fit in the block.

Also adds a fade-out at the bottom of the headings wrapper to indicate that there's more content that can be scrolled into view.

image

## TESTING
 
This code is available on the develop site. If you edit the home page at https://wikimediafoundation-org-develop.go-vip.co/wp-admin/post.php?post=10&action=edit and add additional headings to the hero block they should all be visible.

### DIFF
--- a/assets/src/editor/blocks/home-page-hero/style.scss
+++ b/assets/src/editor/blocks/home-page-hero/style.scss
@@ -22,9 +22,12 @@
 		}
 
 		&__heading-wrapper {
-			z-index: 1;
-			position: relative;
+			height: 100%;
+			mask: linear-gradient(to bottom, #000 90%, #0000 100%) repeat-x;
+			overflow-y: scroll;
 			padding-top: 4rem;
+			position: relative;
+			z-index: 1;
 		}
 
 		&__heading {


MiguelAxcar (approved):
@goldenapples, when I initially took on this ticket, I noticed that interface issue on editor. I invested significant time attempting to make it work locally, even wondering if I had stumbled upon a bug in the process.

For reference, my idea for this UI challenge was to add controls for the multiple headers on the toolbar, possibly incorporating a rotating effect within the editor, allowing the user to switch headers manually, seeing it changing. Also, to implement a progress bar so the user can visually track the rotation over time.

Your current feature is simple and functioning very well, and I would go for adding this improvement even if the original feature isn't fully operational - when can check that with Greg, and potentially file a new ticket.

Btw, I took the liberty of adding you to ticket #891, and moved it to the "Review/QA" pipeline :)

image

varnent (approved):
Works! :)

PR #89, opened 2023-09-26
Update some npm dependencies and clarify install process in README

kadamwhite:
@varnent I spun this up earlier, it will let us make a few of the changes you wanted from the reverted commit without any build impact.

Regarding any Dependabot or `npm audit` security issues: They are all in dependencies of our build process, so none of the specific issues I saw in the report are actual threat surfaces for the theme because we do not run any public-facing node processes (the distribution version of the theme has no dependency code and only includes the built versions of the theme-specific source). See this article for more details, we're squarely in the situation described: https://overreacted.io/npm-audit-broken-by-design/

We can address any such errors when and if we update the supported node version of the build system, and we absolutely aim for a clean bill of health from all auditing tools, regardless of validity — but I recommend they're not urgent under these circumstances.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,9 @@ Begins watching front-end assets (scripts and styles) and compiles them when cha
 * `npm run download-style-guide-sass`
 Download the variables file in the style-guide. This is only required when the style guide has changed.
 
+> [!IMPORTANT]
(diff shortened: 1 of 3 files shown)


varnent (approved):
Thank you @kadamwhite!

PR #90, opened 2023-09-26
88: Fix JS error blocking execution of unrelated plugin code.

### LINKED ISSUE #88: JS error breaks other plugin JS
Relates to:  https://github.com/humanmade/Wikimedia/issues/897

Error output from @kadamwhite: 

```
10 Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')
    at HTMLDocument.<anonymous> (373)
```

The above error is encountered on https://wikimediafoundation-org-preprod.go-vip.net/about/transparency/2022-2/user/ while trying to view Vega charts. Because of VIP script concatenation, an error in one part of the theme’s JS can cause other plugins and theme code to not run at all — we suspect this is the origin of the issue with chart rendering, the JS file errors before the Vega plugin is able to initialize.

This error may be fixable by adding guards to [post-list-filters.js](https://github.com/wikimedia/shiro-wordpress-theme/blob/57cba5fa1b3afb1a25f047762e4589775bc8f641/assets/src/scripts/post-list-filters.js) so that the news searching JS bindings are not triggered on other sections of the site (or at minimum, that the binding functions abort if the querySelector command returns nothing)

kadamwhite:
The filters we introduced in #55 work well, but the code that manages them still runs on pages outside the News section. Because the filter interface doesn't exist on other pages, these `querySelector` calls returned `null`, and attempts to manipulate those null values fail.

This PR slightly refactors the existing code to add guards around the node queries so that we only run code when the corresponding HTML's on the page.

Fixes #88.

**Testing Instructions**

Visit any page other than the News section. No error resembling `TypeError: Cannot read properties of null (reading 'addEventListener')` should appear.

### DIFF
--- a/assets/src/scripts/post-list-filters.js
+++ b/assets/src/scripts/post-list-filters.js
@@ -1,10 +1,16 @@
 /**
- * Toggles the post list filter container.
+ * Set up event handlers to toggle the News section post list filter container, when present.
  *
- * @param {HTMLElement} filterButton - The button used to toggle the filter container.
- * @param {HTMLElement} filterContainer - The filter container.
+ * @param {HTMLElement|null} filterContainer The node containing the filter form.
  */
-const addFilterToggleHandler = ( filterButton, filterContainer ) => {
+const addFilterToggleHandler = filterContainer => {
+	const filterButton = document.querySelector( '.post-list-filter__toggle' );
+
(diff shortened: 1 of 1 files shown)


varnent (approved):
Thank you @kadamwhite!

PR #108, opened 2023-11-08
Nov2023 lint beautify

varnent:
@kadamwhite - is this type of cleanup worth it?


kadamwhite:
@varnent Yes, in principle, but I think it'd be better to use PHPCBF to apply these changes — some of these look manual, and introduce new errors (wrong type of indentation character, etc). 

I believe we should also always be targeting `main` with PRs, putting this into `develop` could result in these changes being wiped out in the future

varnent:
> @varnent Yes, in principle, but I think it'd be better to use PHPCBF to apply these changes — some of these look manual, and introduce new errors (wrong type of indentation character, etc).
> 
> I believe we should also always be targeting `main` with PRs, putting this into `develop` could result in these changes being wiped out in the future

Good points. I am going to take a different tactic. Feel free to wipe any of the effort - mostly exploring for personal pro development sake.

PR #112, opened 2023-11-08
916: Fix PHP fatal: only count team name if it is countable

kadamwhite:
My game of whac-a-mole with PHP 8.1 continues.


MiguelAxcar (approved):
You are close to beat the Whac-A-Mole! :smile:

PR #114, opened 2023-11-08
916: Guard against non-array stories meta

kadamwhite:
Observed on https://wikimediafoundation.org/about/annualreport/2019-annual-report/previous-annual-reports/

For humanmade/wikimedia#916

### DIFF
--- a/template-parts/page/page-stories.php
+++ b/template-parts/page/page-stories.php
@@ -9,6 +9,10 @@
 
 $template_args = get_post_meta( get_the_ID(), 'stories', true );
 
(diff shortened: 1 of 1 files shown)


sacmp:
The same error persists on the Preproduction site, specifically affecting the [Stories Page](https://wikimediafoundation-org-preprod.go-vip.net/stories/) (a section of the Transparency Report). On the front end, the page experiences a breakdown, resulting in some missing content. I've encountered this last week when working on the Transparency Report. I was only able to locate this error on this particular page and couldn't replicate it on other pages.

PR #120, opened 2023-11-21
920: Guard against direct file access

kadamwhite:
VIP runtime logs reveal that some requests are made directly to theme files, which is possibly benign but also possibly as a penetration testing tactic. The file will not load if WordPress is not running, because `get_header()` will not be defined, but this guard code ensures that the page exits immediately if any top-level theme file is accessed.

Fixes humanmade/wikimedia#920

### DIFF
--- a/404.php
+++ b/404.php
@@ -7,6 +7,10 @@
  * @package shiro
  */
 
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // File should never be accessed directly.
+}
+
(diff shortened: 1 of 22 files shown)


kadamwhite:
@MiguelAxcar I'm going to ignore the PHPCS errors on this, since Greg was doing a separate effort to clean up PHPCS in the site.

varnent (approved):

PR #124, opened 2023-11-27
[WIP] Fix PHP warning fixes

MiguelAxcar:
This PR is part of our ongoing effort to maintain clean and efficient logs by addressing and resolving specific PHP warning messages, as keeping the log free of not so important warnings is crucial for efficient debugging and monitoring, ensuring that we can quickly identify and address more significant issues when they arise.

### Changes

1. **sidebar-list.php**:
   - Refined the definition of `$item_text` to default to an empty string if neither 'title' nor 'name' keys are set in `$list_section`, addressing the warning: `PHP Warning: Undefined array key "name" in sidebar-list.php on line 61`.

2. **single-profile.php**:
   - Add lines to set default values for `$link['link']` and `$link['title']` if not present in the `$share_links` array, addressing multiple warnings of the form: `PHP Warning: Undefined array key "link" in single-profile.php on line [77, 80, 83, 86]`.

### DIFF
--- a/single-profile.php
+++ b/single-profile.php
@@ -73,6 +73,10 @@ while ( have_posts() ) :
 							?>
 					<div class="link-list mar-right">
 							<?php
+							// Avoid undefined array index warnings.
+							$link['link']  = $link['link'] ?? '';
+							$link['title'] = $link['title'] ?? '';
+
 							$img = '';
 							if ( is_int( strpos( $link['link'], 'meta.wikimedia.org' ) ) ) {
 								$img = get_template_directory_uri() . '/assets/src/svg/globe.svg';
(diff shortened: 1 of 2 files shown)


single-profile.php
```diff
@@ -73,6 +73,10 @@
 							?>
 					<div class="link-list mar-right">
 							<?php
+							// Avoid undefined array index warnings.
+							$link['link']  = $link['link'] ?? '';
+							$link['title'] = $link['title'] ?? '';
```
kadamwhite, line 78: This patch looks good, but if you have bandwidth, could you try to test out the different profile fields using a recent Production DB dump, and see if you can figure out what data in the CMS causes this error? This patch is a bandaid, and we should know whether the data entry form for profiles is broken in some way that causes this.

PR #125, opened 2023-12-01
Enable media fallback for local dev environments

kadamwhite:
If you [enable a media fallback URL](https://docs.wpvip.com/how-tos/dev-env-add-media/#h-proxy-media-files) to proxy media file requests to the production site (which is helpful because it lets you run your local without downloading the `uploads/` directory!), currently those images do not load because of our CSP. "self" represents the foundation site in production, so it should always be safe to load images from it in other environments, because those environments' data comes one-way from that production foundation site.

With this change, if you create a local environment with a media redirect fallback domain, like this,

```
vip dev-env create --slug=wikimedia --media-redirect-domain=wikimediafoundation.org
```

then all images will load as soon as you run

```
vip dev-env sync sql @wikimediafoundation.production --slug=wikimedia
```

and you will not need to do any manual download of the `uploads/` dir.


kadamwhite:
@MiguelAxcar @goldenapples I noticed that our CSP was preventing this little quality-of-life enhancement that the VIP dev server lets us use, and would propose that we make this change.

Alternatively we can wait for the security plugin, which I'm setting up to enable filtering, so that a local-dev-only plugin could make this change and ensure that the list on prod is exactly what Wikimedia Security wants it to be.

Would welcome y'all's thoughts

PR #129, opened 2023-12-11
Hotfix: PHP Fatal on inc/template-functions.php line 99

MiguelAxcar:
This PR intends to add a hotfix which was breaking role pages on `develop`.

Example of role page:
https://wikimediafoundation-org-develop.go-vip.co/role/board/

Error being thrown:
```
PHP message: PHP Fatal error: Uncaught TypeError: Illegal offset type in /var/www/wp-content/themes/shiro/inc/template-functions.php:99
```

I already deployed this hotfix to Shiro's `develop`, and updated WMF `develop`, fixing the issue, assuring that the fix was effective.

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -104,6 +104,8 @@ function wmf_get_role_hierarchy( int $parent_id ) {
 	foreach ( $terms as $term_id => $parent ) {
 		if ( is_int( $parent ) && $parent > 0 ) {
 			$children[ $parent ][] = $term_id;
+		} else {
+			error_log( sprintf( 'Term ID %d has an invalid parent ID of [%s].', $term_id, var_export( $parent, true ) ) );
 		}
 	}
 
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -93,7 +93,13 @@ function wmf_get_header_cta_button_class() {
 function wmf_get_role_hierarchy( int $parent_id ) {
 	$children   = array();
 	$term_array = array();
-	$terms      = get_terms( 'role' );
+	$terms      = get_terms(
+		'role', array(
+			'orderby' => 'name',
+			'fields'  => 'id=>parent',
+			'get'     => 'all',
+		)
+	);
 
 	foreach ( $terms as $term_id => $parent ) {
 		// Ensure that $parent is an integer and greater than 0.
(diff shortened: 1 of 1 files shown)


inc/template-functions.php
```diff
@@ -86,10 +86,11 @@ function wmf_get_header_cta_button_class() {
...
  * Get all the child terms for a parent organized by hierarchy
  *
- * @param int $parent_id ID to query against.
- * @return array List of organized IDs.
+ * @param int $parent_id The ID of the parent term to query against.
+ * @return array|false An associative array of child terms organized by their IDs,
+ *                     or false if no children are found for the given parent ID.
  */
-function wmf_get_role_hierarchy( $parent_id ) {
+function wmf_get_role_hierarchy( int $parent_id ) {
```
kadamwhite, line 93: If we're adding a type hint here we should also add the `int` hint to `wmf_get_posts_by_child_roles()`, so that we catch any type errors as early as possible.

inc/template-functions.php
```diff
@@ -101,7 +102,8 @@ function wmf_get_role_hierarchy( $parent_id ) {
 	);
 
 	foreach ( $terms as $term_id => $parent ) {
-		if ( 0 < $parent ) {
+		// Ensure that $parent is an integer and greater than 0.
+		if ( is_int( $parent ) && $parent > 0 ) {
```
kadamwhite, line 106: I think we can actually drop this comment, it describes the line of code exactly which feels duplicative. I don't believe this logic would be confusing without the comment.

kadamwhite (changes requested):

inc/template-functions.php
```diff
@@ -101,8 +102,10 @@ function wmf_get_role_hierarchy( $parent_id ) {
 	);
 
 	foreach ( $terms as $term_id => $parent ) {
-		if ( 0 < $parent ) {
+		if ( is_int( $parent ) && $parent > 0 ) {
 			$children[ $parent ][] = $term_id;
+		} else {
+			error_log( sprintf( 'Term ID %d has an invalid parent ID of [%s].', $term_id, var_export( $parent, true ) ) ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
```
kadamwhite, line 108: Ignore needs to be on the line above

kadamwhite (approved):
Approving because the code looks good, but the checker is still reporting an error. Once that is passing you can merge it, but please don't merge it before we get that little green checkmark

inc/template-functions.php
MiguelAxcar: I added this error log to have some info about what is exactly causing the issue, as a way to uncover the root cause.

PR #132, opened 2023-12-11
Remove error log debug info

MiguelAxcar:
A PHP Fatal error was encountered on the Roles listing page. To immediately mitigate this, a provisional fix was introduced in PR [#129](https://github.com/wikimedia/shiro-wordpress-theme/pull/129).

It was needed to trace the root cause of this error, as it may reveal broader systemic issues within our codebase. 

After deploying the fix code, the error log provided us some important info, so logging the error is no longer necessary.

Details added to error log
```
Term ID 990: Invalid parent ID of [0].
Term ID 16: Invalid parent ID of [0].
Term ID 9: Invalid parent ID of [0].
Term ID 10: Invalid parent ID of [0].
```

So, the guard of checking `if ( is_int( $parent ) && $parent > 0 ) {` is enough for the case.

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -104,9 +104,6 @@ function wmf_get_role_hierarchy( int $parent_id ) {
 	foreach ( $terms as $term_id => $parent ) {
 		if ( is_int( $parent ) && $parent > 0 ) {
 			$children[ $parent ][] = $term_id;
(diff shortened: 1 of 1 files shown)


kadamwhite (approved):
Clever, and it all makes sense now!

PR #135, opened 2024-01-11
[#371] Re-enable clock block

kadamwhite:
This PR finally closes the chapter of the mysterious affair of the oft-discussed clock block, which was added to the site, removed, not actually removed, re-added, and then unintentionally disabled by removing the relevant blocks from this theme's allowed blocks list. In this branch we restore the block to the theme's allowed blocks list, and refactor the block so that the script loads conditionally to improve site performance.

### Testing instructions

- On a post or page, insert a "Clock" block; block should be added and show placeholders for title field
- Edit the title of the block inline within the editor
- Edit the time of the block (times in past will count up, times in future will count down)
- Save or publish the block
- View the frontend of the site
- Block should display as expected, and count up appropriately

### Screenshots

Editor:

image

Frontend:

image

### Future enhancements

TODO: Translation support. Currently this is hard-coded to English.


MiguelAxcar (approved):
This change looks very good to me, and I'm glad we're _finally_ releasing the clock block. Great work @kadamwhite .

PR #137, opened 2024-02-06
954: Avoid meta sort in profile query to ensure all records are returned

kadamwhite:
We had an unusual situation on WIKI-954 where the profile of Stephen LaPorte, the Foundation's current General Counsel, was not showing up in the list of profiles with the 'executive' role. Stephen's profile had the correct 'executive' role term associated, and was properly published, but his profile's ID was not returned in the results of the WP_Query from the `wmf_get_role_posts()` function.

I identified that ordering by meta_value causes WP_Query to generate a SQL query which _excludes_ any records which do not have the specified meta key set. This is actually tracked as a long-standing WP Core bug:

https://core.trac.wordpress.org/ticket/19653

The quickest solution to the problem is to remove the `orderby` entirely, and sort the array in-memory within PHP before returning the query results for caching.

I also took it as an excuse to introduce this theme's first PHP Arrow Function :)

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -130,10 +130,6 @@ function wmf_get_role_posts( $term_id ) {
 		array(
 			'post_type'      => 'profile',
 			'fields'         => 'ids',
-			'orderby'        => 'title',
-			'meta_key'       => 'last_name',
-			'orderby'        => 'meta_value',
-			'order'          => 'ASC',
 			'posts_per_page' => 100,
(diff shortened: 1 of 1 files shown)


kadamwhite:
@MiguelAxcar I've used VIP's CLI to set the `last_name` and `profile_role` meta values of Stephen's profile object, and then deleted the cached value. So he's showing again now, but this patch will fix it long-term.

MiguelAxcar (approved):
This change looks good and works well in my local environment.

PR #138, opened 2024-02-07
WMF #956: Remove logic to make Field Manager fields to load only when content has no blocks

MiguelAxcar:
This PR removes a legacy code which was preventing the Field Manager fields to be available when editing posts when the post has blocks. 

This fixes the issue were Field Manager fields were missing when editing profiles.

### DIFF
--- a/inc/fields.php
+++ b/inc/fields.php
@@ -269,28 +269,24 @@ function wmf_get_categories_options() {
 	return $category_list;
 }
 
-// As soon as a post has blocks we should no longer register these fields.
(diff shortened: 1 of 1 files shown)


goldenapples (commented):
I think that many of these Fieldmanager fields actually aren't functional any more for content using the block editor, so we probably don't need to re-enable all of them. Can you just identify the fields that need to be enabled on block content and move them outside of the conditional rather than re-adding all the legacy fieldmanager fields?

kadamwhite (approved):
Delightful.

PR #139, opened 2024-03-07
Remove the Homepage Hero srcset Attribute

ajvillegas:
This change removes the homepage hero `srcset` attribute from the image by removing the `imageId` attribute from the `ImagePicker` component. This prevents the component from generating the `wp-image-{ID}` class needed for WP to automatically generate the `srcset` and `sizes` attributes.

This will allow editors to force the banner to always display the same image regardless of screen size or device in order to prevent blurry images in Chrome.

**Testing Instructions:**
1. Navigate to the demo page here: https://wikimediafoundation-org-develop.go-vip.co/
2. Check that the homepage hero image does not include the `wp-image-{ID}` and that the `srcset` and `sizes` attributes are missing.

### DIFF
--- a/assets/src/editor/blocks/home-page-hero/edit.js
+++ b/assets/src/editor/blocks/home-page-hero/edit.js
@@ -70,7 +70,6 @@ const HomePageHeroBlock = ( { attributes, setAttributes, isSelected, setHeadingC
 				) }>
 					<ImagePicker
 						className="hero-home__image"
-						id={ imageId }
 						labels={ {
 							instructions: __( '⚠️ Scrolling images are expected to be 4237px high and 454px wide (or multiples of those values). Using a different resolution will result in strange behavior.', 'shiro-admin' ),
 						} }
--- a/assets/src/editor/blocks/banner/index.js
+++ b/assets/src/editor/blocks/banner/index.js
@@ -221,6 +221,7 @@ export const settings = {
 					<ImagePicker.Content
 						alt={ imageAlt }
 						className="banner__image"
+						id={ imageID }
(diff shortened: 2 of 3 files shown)


joeleenk (commented):
Hmm, I'm concerned about how this change will affect mobile users. Would it be possible to add another image size to the srcset to target Chrome and make sure the image is cropped to a max size, instead of trying to load the original image?

joeleenk (approved):

joeleenk:
Approving, as this is time sensitive, and this fix is an improvement on what we currently have. We're going to open a ticket for exploring a more robust solution in a future allocation.

joeleenk (approved):

assets/src/editor/blocks/home-page-hero/edit.js
```diff
@@ -70,7 +70,6 @@ const HomePageHeroBlock = ( { attributes, setAttributes, isSelected, setHeadingC
 				) }>
 					<ImagePicker
 						className="hero-home__image"
-						id={ imageId }
```
goldenapples, line 73: Does this attribute need to be removed here? Seems like it might be useful to continue storing the id attribute, even if it's not output in the save component.

goldenapples (approved):
Sorry for missing this earlier. This is ok to merge, I just had a question about whether to leave the id attribute in the edit component.

assets/src/editor/blocks/home-page-hero/edit.js
ajvillegas: I removed it to prevent the unmatched output error, but I did not fully tested this since I was not able to setup my local env.

PR #143, opened 2024-04-02
Report plugin integration: Permit page title suppression

kadamwhite:
This allows other parts of the site to choose to disable the template-side page title if they provide it in post content


mattwatsoncodes (approved):
LGTM 🥳

PR #149, opened 2024-05-27
894: URL Editor Icon Does Not Display for Card Block

ajvillegas:
This change fixes the bug where the URL icon for editing the CTA link in the Card block was not showing up on focus.

For https://github.com/humanmade/Wikimedia/issues/894.


kadamwhite (approved):
Well spotted, @ajvillegas

PR #154, opened 2024-06-27
1020: Arrange profiles specific orders

pamprn09:
Since we are not always using the role taxonomy, I've added a Custom Field on the Profile CPT to set the order.
The post is sorted at the post list according to its Profile order.


kadamwhite:
Superseded by #155 after discussion with @pamprn09

PR #155, opened 2024-06-28
#1020: Fix profile ordering on staff page and support manual role ordering

kadamwhite:
The key change here is to `usort` all profiles by `last_name` meta value (which has already been coopted for sorting purposes), but importantly to treat a missing value as `z` instead of `''`. Maryana Iskander's profile already had the `last_name` key `1Iskander`, but this wasn't displaying first because (a) that section wasn't named in a way that it triggered any of the hard-coded logic in the template, and (b) Margo Lee had _no_ `last_name` set and an empty string apparently comes before `'1'` when PHP compares by `strnatcasecmp`.

To control sort order on the staff page:

1. Make sure the "Sort Name" (formerly "last name") metabox on each employee profile is filled out with values that will sort correctly, numbers before letters and letters in alphabetical order.
2. Make sure the categories themselves have a proper `role_order` value set, to make them display in the right order on the staff page. For example, the Office of the CEO could be '1', Communications could be '2', Advancement '3', etc.
    - Child categories can be ordered by adding decimal numbers, so to put "External Comms" before "Brand", you'd give External Comms `1.1`, Brand `1.2`, etcetera

### DIFF
--- a/inc/fields/profile.php
+++ b/inc/fields/profile.php
@@ -240,9 +240,14 @@ add_action( 'update_postmeta', 'wmf_save_user_profile', 10, 4 );
  */
 function wmf_render_role_order_list_table_column( ?string $string, string $column_name, int $term_id ) {
 	if ( $column_name === 'role_order' ) {
+		$parent_term_id = wp_get_term_taxonomy_parent_id( $term_id, 'role' );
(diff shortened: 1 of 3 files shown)


pamprn09 (approved):
Looks good!

PR #156, opened 2024-07-29
1012: Modernize Shiro frontend build to use wp-scripts

kadamwhite:
This is a big PR, but it performs a lot of parallel and overdue actions.

- Unifies the two different build pipelines
  - Modernizes the HM-Webpack-helpers-based Webpack build
  - Sunsets the legacy Gulp build
  - Replaces most of the custom functionality — RTL etcetera — with out-of-the-box `wp-scripts` functionality
  - Uses `wp-scripts build` to build all code, through the introduction of a `webpack.config.js` designed to adapt the default WP scripts build behavior to match how the previous build named files. This allows us to avoid _most_ source file changes.
- Updates source files when necessary
  - Modernizes and modularizes SASS import usage
  - Adjust `url()` paths to be relative to `assets/dist`, because we use a CSS file in there now instead of the top-level `style.css` (which means the theme is now a valid WP theme upon clone, and can be activated from `main` without running a build)
  - Adjusts the entrypoints for the editor sass bundles
- Moves a doesn't-change-much SVG spritesheet build into a committed file to speed up prod builds
- Simplifies build dependency chain where possible, dramatically reduces Package.json dependencies
- Switches project to use `wp-scripts`-derived ESLint configuration
- Uses `wp-scripts lint` to run ESLint
- Updates all builds to use Node 20, which is the new required baseline for most GH Actions

> [!WARNING]
> THIS DOES BREAK HOT RELOADING.
> We can add it back in next time we get to block development, but for now I propose that the value in upgrading the overall build is worth having to reload a little more manually for the next sprint or two.
> (I'm looking into why, since the `--hot` flag is not having any effect even if the Gutenberg plugin is active; WP core docs might be outdated.)


kadamwhite:
Merged to [Develop](https://wikimediafoundation-org-develop.go-vip.co/) for testing, and VIP's concatenation can be disabled to view the specific files that are loading by adding `?concat_css=false` to the end of a URL.

kadamwhite:
Our colleague Fränk suggests using the asset files instead of md5-hashing assets for versioning. This should be done, but I'd vote to do it as part of a later phase; whenever we next work on block code (WIKI-780, perhaps) we can refactor how blocks get initialized and fix up asset hashes at the same time.

allysonsouza (approved):

PR #158, opened 2024-09-18
1037: Fix Block Editor Styles Not Loading Properly

ajvillegas:
This PR adds fixes the block editor styles not loading properly for custom blocks.

For https://github.com/humanmade/Wikimedia/issues/1037


kadamwhite (approved):
Would have loved to see us update this to the [modern way we're supposed to be doing editor styles](https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-support/#editor-styles), but that can come during the refresh.

PR #159, opened 2024-09-24
1037: Fix focus bug related to deprecation of multiline content

kadamwhite:
WordPress 6.2 deprecated the `multiline` attribute of the RichText component in favor of InnerBlocks, which sounds good in theory but they did not leave a clean migration path from the one to the other. I tried everything possible to get a JS-side block deprecation working, and eventually gave up in favor of manually filtering the content when fulfilling a `context="edit"` REST request to adjust the markup into the form the new version of the block expected.

This will fix the immediate issue, and we can script a broader migration later if necessary.

I ended up pursuing a number of red herrings during this process

- Resolved a few JS-side deprecation warnings in the code
- Added a manual control to make the Landing Page Hero open in a new tab (helps with avoiding block validation errors)
- Remove keepPlaceholderOnFocus properties on RichText, which were throwing warnings in the JS console and do not (from my testing) appear to impact the actual functionality of the component any longer...?
- Fix a deprecation warning around using the wrong filter for `block_categories_all`
- Fix a PHP warning thrown if running this theme outside VIP (I was trying everything I could think of while attempting to trace the origin here...)

This PR builds on top of #158.


goldenapples (approved):
I don't have a local set up to test this, but the code looks good.

Out of curiosity, was this approach (filtering the post content in PHP when sending to the editor) simpler than applying the transformations in the block's `edit()` component?  Some of these transformations, especially the regex matching and building block markup through string replacement, seem like they'd be more suited for scripting in JS.

PR #162, opened 2024-10-08
#1017 - Set width on alignfull core-group blocks

allysonsouza:
Set width on `alignfull` core-group blocks.

### DIFF
--- a/assets/src/sass/helpers/mixins/_mixins-gutenberg.scss
+++ b/assets/src/sass/helpers/mixins/_mixins-gutenberg.scss
@@ -3,6 +3,7 @@
 	margin-left: rem(-20);
 	margin-right: rem(-20);
 	min-width: 100%;
+	width: 100vw;
 
 	@media ( min-width: #{$width-breakpoint-tablet} ) {
 		margin-left: rem(-40);


assets/src/sass/base/_base.scss
```diff
@@ -1,17 +1,74 @@
...
+@media screen and (max-width: #{$width-breakpoint-tablet}) {
+	:root {
+		--wp--style--global--content-size: calc( 100vw - 40px );
+		--wp--style--global--wide-size: calc( 100vw - 40px );
+	}
+	.single-post {
+		--wp--style--global--content-size: calc( 100vw - 40px );
+		--wp--style--global--wide-size: calc( 100vw - 40px );
+	}
+}
```
kadamwhite, line 50: There may well be a better way to handle this, but using media rules to alter these variables was the simplest way I could achieve today's padding behavior without actually adding complex padding rules.

assets/src/sass/base/_base.scss
```diff
@@ -1,17 +1,74 @@
...
+
+.is-layout-constrained > :where(:not(.alignleft):not(.alignright):not(.alignfull)),
+// For now, top-level groups with background colors also behave as layout-constrained.
+.page .article-main > .has-background.alignfull > :where(:not(.alignleft):not(.alignright):not(.alignfull)) {
+	max-width: var(--wp--style--global--content-size);
+	margin-left: auto !important;
+	margin-right: auto !important;
+}
```
kadamwhite, line 58: The styles here and below mirror the latest from core / twentytwentyfour

assets/src/sass/editor/_global.scss
```diff
@@ -3,101 +3,39 @@ body {
...
-		&[data-align="wide"] {
-			max-width: $content-column-padded-rem;
-
-			.has-background {
-				padding: 1.25em 2.375em;
-			}
-		}
+.has-background > .wp-block-group__inner-container > :last-child {
+	margin-bottom: 0;
+}
```
kadamwhite, line 17: These editor styles remain a bit of a mess, but are simpler than before, and I'm hesitant to further simplify until we have a new target style to implement as part of the refresh project.

assets/src/sass/base/_base.scss
```diff
@@ -1,17 +1,74 @@
...
+
+/* Layout (block style)
+--------------------------------------------- */
+:root {
+	--wp--style--global--content-size: 980px;
+	--wp--style--global--wide-size: 1020px;
+}
+.single-post {
+	--wp--style--global--content-size: 40rem;
+	--wp--style--global--wide-size: 980px;
```
goldenapples, line 27: Oh, this is nice. I didn't realize changing content width could be done this easily. In another project we did it with a filter on "block_editor_settings_all".

Is there a risk in mixing units here? If a user changes the accessibility settings in their browser so that 1rem = 30px, the "content" width might be larger than the "wide" width.

kadamwhite: Interesting point. The content widths we are dealing with (posts content column and overall menu column width) _are_ designated in these exact units, so I think in this case things will behave the most consistently if we leave as-is. But I think we should refactor the overall site widths to either rems or px but not both when we hit the Refresh.

PR #166, opened 2024-10-18
1017 alignment followup

kadamwhite:
Two final changes

### 1. Related Posts should be "wide" width

Before | After
----- | -----
image | image

### 2. Display-Inline buttons in main content column don't obey margins

Before | After
----- | -----
image | image


stuartshields (approved):
@kadamwhite LGTM 🚀

PR #167, opened 2024-10-18
1017: Ensure embedded Blog List blocks render as wide-width in posts

kadamwhite:
Blog List block should display wide-width in posts.

Screenshots **after**:

image

image


stuartshields (approved):
@kadamwhite LGTM 🚀

PR #168, opened 2024-10-31
1043: Modernize block registration patterns within Shiro (Part 1)

kadamwhite:
WIKI-1043

This PR adjusts how we load some of the Shiro blocks to bring our block registration in line with [the recommended `block.json` registration pattern](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-metadata/) introduced in WordPress 5.8.

It represents a cross section of block types, with these initial blocks being available for testing:

- Accordion (basic JS block)
- Accordion Item (basic JS block with editor styles)
- Banner (basic JS block with editor styles)
- Blog List (dynamic block)

I would like to validate all of these blocks work as expected (and that other theme functionality, including blocks which have not yet been migrated, continue to work) before proceeding.

To prepare for this process (and resolve issues during these initial four blocks), these overall changes have been made:

- Decouple block registration from Block HMR Helpers autoloading so that we can gradually move blocks over to their own files
- Update the Webpack build to fix a bug where `block.json`-based blocks were not detected (causing their scripts to not be output)
- Fix an issue in the Webpack configuration where multiple block editor styles would be generated with the same chunk name, breaking the splitChunks plugin's processing
- Introduce PHP code to automatically detect and load `block.json` blocks on the server, reducing the amount of manual boilerplate needed to register a block

For each block, these changes were made:

- Move `attributes` to a `block.json` file
- Update each block entrypoint module to register its own block, instead of exporting block settings for external registration
- [Register the scripts and styles](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-metadata/#editor-script) for a block in its `block.json` file, as `editorScript`, `editorStyle`, etc
- Move any server-side rendering code into a `render.php` file referenced from within `block.json` (see Blog List block for example)

Testing:

These four blocks should all be able to be added, saved, and edited as expected.

image


pamprn09 (approved):
I've tested the blocks and they are working as expected. Nice job modernizing this code!

kadamwhite:
Thanks for the review!

PR #170, opened 2024-11-05
WIKI-1043: Modernize block registration patterns within Shiro (Part 2)

kadamwhite:
WIKI-1043

This PR contains block.json modernization and code cleanup for the following blocks

- blogPostHeading
- card
- carousel
- clock
- clockStat
- collapsibleText

All of these blocks should work as they did before, but attributes should now be defined in block.json, they should not be loaded by default as part of the editor bundle, and any frontend assets should be conditionally loaded through linkage via block.json file: references


assets/src/editor/blocks/clock/frontend.scss
```diff
@@ -50,6 +54,7 @@
         border: rem(2) solid darkgrey;
         display: flex;
         font-weight: bold;
+				min-height: rem(19);
```
allysonsouza, line 57: Indentation

allysonsouza (approved):

PR #171, opened 2024-11-05
WIKI-1037: Allow opening banner block CTA in new tab

kadamwhite:
Addresses a remaining piece of feedback around the banner block after WIKI-1037, that the CTA should be able to open in a new tab without manually editing the markup saved in the post.

Adds an `openInNewTab` attribute to the banner block, and passes it through to the CTA, which exposes a control in the sidebar:
image

Since the banner block has many styles and this section may be off-screen and require scrolling, I've also introduced a toolbar button to make it more obvious:
image

In action:
banner-cta-open-in-new-tab

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,5 +14,8 @@ indent_style = tab
 indent_style = space
 indent_size = 2
 
+[block.json]
+indent_style = tab
+
 [composer.json]
 indent_size = 4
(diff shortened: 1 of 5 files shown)


pamprn09 (approved):
I've tested this feat and it's working as expected.

PR #179, opened 2024-12-20
1010: Re-enable dropdown-style share button

kadamwhite:
@pamprn09 Per @sacmp ,

> I recommend not rolling out the share button yet, as I need more time to test its potential impact on the other social media buttons on the website.

We therefore added in these styles and PHP to re-enable the old button method; This reverts commit 8d50dea9649b4255893e2c6e56365f36f9f13b18.

Now that the other version has been deployed, we can proceed with merging this through to preprod.


pamprn09 (approved):

allysonsouza:
I don't know why yet, but this property seems to be building to `display: inline-box`, which is invalid, but it happens with `contents` or other values like `inline-flex` or `inline-block`: https://github.com/wikimedia/shiro-wordpress-theme/blob/1010-re-enable/assets/src/sass/blocks/_share-article.scss#L14

PR #181, opened 2025-01-09
1024: Add query parameter to determine referer to profile page

allysonsouza:
Add a query parameter to profile links to determine which page was the referer to point it back on breadcrumb.

### DIFF
--- a/single-profile.php
+++ b/single-profile.php
@@ -27,7 +27,8 @@ while ( have_posts() ) :
 	if ( ! empty( $roles ) && ! is_wp_error( $roles ) ) {
 		$team_name  = $roles;
 		$ancestors  = get_ancestors( $roles[0]->term_id, 'role' );
-		$referer_id = isset( $_GET['referer'] ) ? absint( $_GET['referer'] ) : null; // phpcs:ignore Processing form data without nonce verification.
+		// phpcs:ignore Processing form data without nonce verification
+		$referer_id = isset( $_GET['referer'] ) ? absint( $_GET['referer'] ) : null;
 		$referer_link = get_term_link( $referer_id );
 
 		// Check if referer query parameter is set to determine $parent_id before lookup for ancestors.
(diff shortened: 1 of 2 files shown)


allysonsouza:
@kadamwhite I don't get how this is different than what I've implemented, do you mean by passing the actual URL into the query parameter instead of the term ID, that way the cache would be bypassed somehow?

kadamwhite:
@allysonsouza To talk through it, think about [VIP's page cache](https://docs.wpvip.com/caching/page-cache/). Imagine this scenario:

Actions | Profile cache state
----- | -----
1: I go to the executive team and click on Maryana's profile | empty
2: Maryana's profile loads, and $_GET['referer'] is the Executive page | Shows "Executive"  
3: You go to the Staff page and click on Maryana's profile | Shows "Executive"
4: VIP loads Maryana's page from cache, showing "Executive", despite where you came from | Shows "Executive"

We cannot rely on $_GET, because that will only be relevant for _uncached_ pageviews. If we reimagine the scenario above but you're browsing between staff categories yourself, (unless you're logged in) you'll see incorrect breadcrumbs even in your own session.

The alternative @goldenapples suggested would be to modify all "role" archive pages (probably via some JS snippet that adds a query string to the links) to send you to the profile with a `?referer=<role slug or ID>` parameter. Then, when rendering the page, if that query parameter is present, we can use that instead of the default parent. This works exactly like the $_GET approach, _except_ that because it's a different URL it won't be the same cache key on VIP's side. `profiles/maryana`, `profiles/maryana?referer=staff`, and `profiles/maryana?referer=executive` (simplified for example) should all be cached separately.

allysonsouza:
@kadamwhite I'm sorry, I still don't understand, I'm doing exactly that, adding query parameters to the profile links with `add_query_arg()` and then using the `$_GET['referer']` only to fetch from which page the user came from, so we do have different url's for each profile (but I'm using ID instead of role slug), isn't that enough to make the cache consider they are different pages in this scenario?

kadamwhite (approved):
After realizing that there are changes to `role-item.php` to add a _custom_ referer parameter, my complaint is entirely unfounded 🤦 

Add a `// phpcs:ignore` for the nonce verification warning, and this looks like it should do the trick

PR #182, opened 2025-01-20
1034: Remove unused CSS

allysonsouza:
Cleanup CSS files looking for:

- Commented properties
- Browser specific properties and prefix that aren't needed anymore
- Removing clearly unused CSS rules
- Empty selectors and "TO-DOs" from years ago
- Unused mixins


goldenapples:
So far it removes a little under 10k from each stylesheet (style.css from 348K to 339K, and editor-style.css from 423K to 416K). Decent start and it looks like you're removing styles in a logical way. I'd hope we can still cut out a lot more of these styles (I was hoping, seeing the scope of the ticket, that we could cut the total weight of css over the pipe by 20%), but whatever state we can get it to by the end of the allocation is an achievement.

allysonsouza:
I ended up just reading trough the `.scss` files and getting what is not being used, commented sections, searched in editor for mixins, to check if they're being used or not, and remove the one's that aren't in use, so that's the methodology adopted.

I was thinking that maybe we can get some kilobytes by tuning the build, I don't know if comments are stripped out by default, for example. And for readability and maintainability, I would format every `.scss`too.

allysonsouza:
But I would do the formatting on a separate PR, doing now would mess the visualization of removals.

goldenapples (approved):
I see nothing risky in here, seems good to merge!

Before closing the ticket we should write up some followups, and I'd leave the issue for @kadamwhite to review, but this can be the last deploy of the print if you can get it through...

allysonsouza:
@goldenapples unfortunately I'm seeing a minimal issue on my local, in the block group/columns, I don't if it's related, I'm updating my local content to have the post images and see if still looks broken to be sure. It would be safe to merge it and see on dev if it's working?

goldenapples:
Yeah, dev is a developer playground, you can merge to dev at will for testing and review.

sacmp:
Thank you, @allysonsouza. I tested the removal of unused CSS and didn’t encounter any issues. Could you share specific testing recommendations to ensure I haven’t overlooked anything? cc @goldenapples

PR #185, opened 2025-02-21
1010: Update Share Button

ajvillegas:
This change updates the share button with the following:
* Update share button horizontal alignment and positioning within the single page template.
* Update the alignment for the dropdown so all icons align with the main share icon.
* Enable all sharing options to the share button in the header and footer page template.
* Add Bluesky sharing option.
* Add a block template part of `article-share` for updating the share button option hardcoded within the header and footer templates for single articles.


goldenapples (approved):
Looks good to me.

ajvillegas:
I [removed the Page Migration Flow sidebar plugin](https://github.com/wikimedia/shiro-wordpress-theme/pull/185/commits/2f8abe134b165f924b74057bf4145505fd7aa97d) as it was preventing us from saving the share-article block template part that was added in this PR.

goldenapples (approved):
The code all looks good. I thought that part of the feature was to remove X/Twitter from the default share icons... should that be handled here in the theme, or is is it necessary to make it a site-level override?

ajvillegas:
@goldenapples Do we need to completely remove Twitter (X) as an option? If so, then yes that should be handled in this PR as well.

I'll reach out to @sacmp for clarification.

sacmp:
Hi @ajvillegas thank you so much for this amazing work! Because you were able to find a way to show/hide the X/Twitter social button in the header and footer, I believe we don't need to remove it at all from the website. We just needed a way to hide it for now. Thank you Alexis, I think this can go to preproduction for testing!

PR #196, opened 2025-05-23
1088: Donation bar pattern and form block

roborourke:
- **#1088 Donation Bar pattern and form block**
- **Add pattern file**

https://github.com/user-attachments/assets/272eedae-e6bd-454d-8551-b2b997b77475

https://github.com/user-attachments/assets/76968226-fc0a-4142-8637-7cbcc646838a

### DIFF
--- a/assets/src/editor/blocks/donation-bar/index.css
+++ b/assets/src/editor/blocks/donation-bar/index.css
@@ -32,6 +32,11 @@
     border-radius: 4px;
     box-shadow: 0 0 1px rgba(0, 0, 0, .2);
     font-size: 1rem;
+
+    &:focus-within {
+        outline: 2px solid var(--wp--preset--color--base);
+        outline-offset: 2px;
+    }
 }
 
 .wp-block-shiro-donation-portal-form__amount abbr {
@@ -52,6 +57,7 @@
     border-radius: 0 4px 4px 0;
     border: 0;
     margin: 0;
+    outline: none;
     padding: calc( 2px + var(--wp--preset--spacing--30) ) var(--wp--preset--spacing--30);
     width: 6rem;
 }
@@ -70,7 +76,6 @@
     border: 2px solid currentColor;
     border-radius: 50%;
     background: transparent;
-    outline: none;
     display: inline-block;
     position: relative;
     transition: border-color 0.2s ease-in;
@@ -84,6 +89,7 @@
 
     &::before {
         content: '';
+        cursor: pointer;
         display: block;
         width: 75%;
         height: 75%;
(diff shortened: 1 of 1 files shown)


assets/src/editor/blocks/donation-bar/index.css
```diff
@@ -0,0 +1,117 @@
...
+    padding: 0;
+    display: flex;
+    gap: var(--wp--preset--spacing--40);
+}
+
+.wp-block-shiro-donation-portal-form legend {
+    position: absolute;
+    width: 1px;
+    height: 1px;
+    visibility: hidden;
```
kadamwhite, line 20: Could we have piggybacked on a global .screen-reader-text here...?

assets/src/editor/blocks/donation-bar/index.js
```diff
@@ -0,0 +1,254 @@
...
+			</>
+		);
+	},
+
+	/**
+	 * Save the share article block, it's a dynamic block.
+	 */
+	save: function Save( { attributes } ) {
+		return null;
+	},
```
kadamwhite, line 245: If the save method is only to return null, we can omit the save method entirely during registration and it'll naturally fall back to the PHP render

assets/src/editor/blocks/donation-bar/index.js
```diff
@@ -0,0 +1,254 @@
...
+	SelectControl
+} from '@wordpress/components';
+
+import metadata from './block.json';
+
+import './index.css';
+
+const supportedCurrencies = [
+	{ label: 'USD ($)', value: 'USD' },
+	{ label: 'CAD (CA$)', value: 'CAD' },
```
kadamwhite, line 29: Could we localize this array from PHP, since we define it in geoip on the backend? Feels like a foot gun to have to maintain it two places

kadamwhite (approved):
Not in a position to test locally, right now, but the logic makes sense and this is an interesting way to handle it. Took a while to understand exactly why we need to maintain an option with a list of paths, that could be worth a little more documentation (in code or in a geolocation approach wiki page).

One note about duplicating the currency array, everything else was just nitpicks :) 🚢

assets/src/editor/blocks/donation-bar/index.css
roborourke: Sara showed us some future designs on the standup yesterday where these would be displayed, so I figured we can add a block style for those when the time comes.

assets/src/editor/blocks/donation-bar/index.js
roborourke: Updated, cheers for the push.

roborourke:
I linked to the VIP docs at the top of the geoip file, and added the `_doing_it_wrong()` check but an extra comment could be handy for sure.

roborourke:
@joeleenk i think Sara is happy for us to merge this one and close the ticket if it helps

PR #203, opened 2025-06-03
Improve DevServer behavior

kadamwhite:
- Merges #170 into the UX Refresh branch (it was meant to be merged quite a while back, and updates a lot of the block registrations to use the modern per-block modules pattern)
- Updates the hot blocks helpers to provide better support for handling filters, if present
- Reloads DevServer on `theme.json` change
- Suppresses noisy DevServer error overlay which can't be disabled due to internal DevServer logic bug
- Remove now-unneeded Composer patch which makes composer install take up till `timeout` while installing
- Updates build dependencies, which seems to resolve some issues
- Fixes or suppresses remaining SASS deprecation warnings so that real issues can be more easily seen in Webpack output
- Moves editor.js registration to wp-scripts "asset.php" loading paradigm
- Dynamically sets publicPath for hot reloading chunks so that we can hot-reload within the editor.js bundle
- Turns off the block directory so that the site doesn't try to dynamically look for missing blocks on the open web


joeleenk (approved):
Thanks for helping to debug this! I did more of a cursory review, since it's been reviewed already in the original PR https://github.com/wikimedia/shiro-wordpress-theme/pull/170, but everything looks great - the errors are gone, and the shiro blocks load in webpack dev server mode. 🎉

goldenapples (commented):
It's got a merge conflict right now, but the original PR looks good. Do you want another review once the merge conflict is fixed?

PR #205, opened 2025-06-03
1117: UX Refresh - Update category slug classes for styles

joeleenk:
I didn't initially realize that some of the categories exist / we might be able to keep some existing category slugs and rename them. This PR is to align the new styles with what the slugs end up being.

https://github.com/humanmade/Wikimedia/issues/1117

### DIFF
--- a/assets/src/sass/helpers/mixins/_mixins-links.scss
+++ b/assets/src/sass/helpers/mixins/_mixins-links.scss
@@ -154,7 +154,7 @@
 	&.category-technology::before {
 		background-color: var(--wp--preset--color--bright-blue);
 	}
-	&.category-policy::before {
+	&.category-public-policy::before {
 		background-color: var(--wp--preset--color--bright-green);
 	}
 	&.category-community::before {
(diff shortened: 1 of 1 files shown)


assets/src/sass/helpers/mixins/_mixins-links.scss
```diff
@@ -157,7 +157,7 @@
 	&.category-public-policy::before {
 		background-color: var(--wp--preset--color--bright-green);
 	}
-	&.category-our-community::before {
+	&.category-community::before {
```
kadamwhite, line 160: Would it help the migration period if we supported both classes? Or is only the new class valid?

kadamwhite (approved):

assets/src/sass/helpers/mixins/_mixins-links.scss
joeleenk: No, I actually created the "our-community" one before I realized the other one existed. And for the migration period, worst case is that the dots are all blue instead of multi-colored.

PR #207, opened 2025-06-04
1103: UX refresh - fix color contrast values

joeleenk:
- Add text color for new colors that should have reverse text
- Remove new colors from old list to prevent overlapping styling
- Remove button class from automated text colors - these will be addressed with button styles
- Add important to link styles for `.wp-block-shiro-button.is-style-as-link` for old colors, so that old instances don't change
- Filter the icon color for shiro-button link for new colors

https://github.com/humanmade/Wikimedia/issues/1103

### DIFF
--- a/assets/src/sass/utilities/_theme-supports.scss
+++ b/assets/src/sass/utilities/_theme-supports.scss
@@ -1,7 +1,25 @@
 // Definitions of editor styles registered as theme supports.
 
+// Support accessible color combinations for background colors.
+$color-palette-reverse:
+	"blue",
+	"blue-aaa",
+	"green",
+	"green-aaa",
+	"purple",
+	"red",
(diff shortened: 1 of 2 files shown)


assets/src/sass/utilities/_theme-supports.scss
```diff
@@ -61,12 +72,13 @@ $color-palette:
 		color: $text-color;
 		color: var(--text-color);
 
-		&.wp-block-button__link,
 		.wp-block-shiro-button.is-style-as-link {
-			color: $link-color;
-			color: var(--link-color);
+			color: $link-color !important;
```
kadamwhite, line 76: Oh, the joys of overly specific core button styles

kadamwhite (approved):
Good disambiguation

PR #214, opened 2025-06-25
1105: Promo carousel edits

joeleenk:
https://github.com/humanmade/Wikimedia/issues/1105

Follows on to https://github.com/wikimedia/shiro-wordpress-theme/pull/206

- Add arrow link button style
- Tweaks to carousel styles
- Create pattern for promo carousel


ajvillegas (approved):
This is all very nice, but why add the pattern through PHP instead of placing it inside the `patterns` folder?

PR #220, opened 2025-07-07
1141: ux refresh bug fixes

joeleenk:
https://github.com/humanmade/Wikimedia/issues/1141

- Fix landing page hero background position on wide screens
- Don't show empty translations in inline-languages list
- Update header donate button styles on mobile to wrap longer text inside the button, instead of breaking the button to a new line

### DIFF
--- a/assets/src/sass/components/_site-header.scss
+++ b/assets/src/sass/components/_site-header.scss
@@ -34,7 +34,11 @@
 	}
 
 	.donate-nav {
-		@include media("<mobile") {
+		@include media("<tablet") {
+			max-width: 36%;
+		}
+
+		@include media("<30rem") {
 			max-width: 24.5%;
 		}
 


inc/editor/blocks/inline-languages.php
```diff
@@ -41,12 +41,21 @@ function register_block(): void {
  * @return string
  */
 function render_block( $attributes ) {
+	$translations = array_filter( wmf_get_translations(), function ( $translation ) {
+		return $translation['uri'] !== '';
```
kadamwhite, line 45: Haven't looked at that function in a while — is there a risk it'd be unset, instead of `''`?

_If so_, this might do the same thing without any risk of undefined array key warnings:

```suggestion
		return ! empty( $translation['uri'] );
```

If it's guaranteed to exist, the check is A-OK as-is.

assets/src/sass/components/_site-header.scss
```diff
@@ -37,7 +37,9 @@
 		.wp-block-button__link {
 			@include media("<mobile") {
 				font-size: rem(16);
-				line-height: var(--wp--custom--line-height--relaxed);
+				margin-block: rem(4);
+				max-width: rem(84);
```
kadamwhite, line 41: 84 feels a little arbitrary, I'm guessing that was just what looks good?

kadamwhite (approved):

inc/editor/blocks/inline-languages.php
joeleenk: Good point, I've updated both of the blocks that I know use this check.

PR #223, opened 2025-07-11
Update Reverse Color Palette Class Rules

ajvillegas:
This change updates the reverse color palette class rules to add rules for all the colors in order to prevent a nested group block from overwriting another child group block with a color background that was not previously accounted for.

### DIFF
--- a/assets/src/sass/utilities/_theme-supports.scss
+++ b/assets/src/sass/utilities/_theme-supports.scss
@@ -75,12 +75,41 @@ $color-palette-reverse-light:
 		.is-style-transparent.has-icon .wp-block-button__link:not(:active)::before,
 		.is-style-as-link.has-icon .wp-block-button__link::before,
 		.is-style-as-arrow-link.has-icon .wp-block-button__link::before {
-			filter: brightness(0) invert(1) !important;
+			filter: brightness(0) !important;
 		}
 
 		.wp-block-button__link {
 			border-color: var(--wp--preset--color--main) !important;
 			outline-color: var(--wp--preset--color--main) !important;
+		}
+
+		&.has-base-background-color .wp-block-button__link {
+			&:hover,
+			&:focus {
+				border-color: currentColor !important;
+
+				&:is(.is-style-secondary a):not(:active) {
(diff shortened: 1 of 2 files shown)


assets/src/sass/utilities/_theme-supports.scss
```diff
@@ -4,8 +4,8 @@
 	filter: brightness(0);
 }
 
-// Support accessible color combinations for background colors.
-$color-palette-reverse:
+// Support accessible color combinations for dark background colors.
+$color-palette-reverse-dark:
```
joeleenk, line 8: I was using "reverse" to mean white text on a dark background, so if we want to switch it to "light" and "dark", it would just be `$color-palette-dark` and `color-palette-light`.

assets/src/sass/utilities/_theme-supports.scss
```diff
@@ -39,6 +39,56 @@ $color-palette-reverse:
...
+@each $color in $color-palette-reverse-light {
+	.has-#{$color}-background-color {
+		color: var(--wp--preset--color--main) !important;
+
+		.wp-block-shiro-button.is-style-as-link::before,
+		.wp-block-shiro-button.is-style-as-arrow-link::before,
+		.is-style-transparent.has-icon .wp-block-button__link:not(:active)::before,
+		.is-style-as-link.has-icon .wp-block-button__link::before,
+		.is-style-as-arrow-link.has-icon .wp-block-button__link::before {
+			filter: brightness(0) invert(1) !important;
```
joeleenk, line 78:
```suggestion
			filter: brightness(0)!important;
```

We want the icons to be black, like the text.

joeleenk (commented):
This works pretty well, just a couple of issues with button styles on white.

ajvillegas:
Thank you both for the review!

@joeleenk Thank you for the code commit. I renamed the variables based on your suggestions.

@kadamwhite Given the code updates from Joeleen the code is different enough that I think it's best to not try and merge both functions, but please let me know if you don't agree.

joeleenk (approved):

PR #232, opened 2025-07-23
1105: Promo carousel UAT edits

joeleenk:
https://github.com/humanmade/Wikimedia/issues/1105

- ~~Return carousel JS early for <2 slides~~
- Carousel column order magic


ajvillegas (approved):
This works great with both the spacer block and the group block with set height.

PR #241, opened 2025-08-18
1126: Add Testimonials Carousel Block Pattern

ajvillegas:
https://github.com/humanmade/Wikimedia/issues/1126

This change adds a Testimonials Carousel block pattern based on the provided design mockups.

Demo of the pattern in Dev (database only; this branch has not been merged with Dev): https://wikimediafoundation-org-develop.go-vip.co/testing-testimonials-pattern/.

Still to do:

- [ ] Update carousel block styles so the arrows sit below each slide on smaller screens without any overlap.
- [ ] Styles for small screens for each individual testimonial card.
    - Could use container queries so it adjust well even before the carousel switches to one column when displaying more than one slide per page.
- [ ] Adjust the button text link color style rules to account for the black color when used within a lightest green color background.

### DIFF
--- a/assets/src/editor/blocks/accordion-item/block.json
+++ b/assets/src/editor/blocks/accordion-item/block.json
@@ -14,7 +14,7 @@
 		"title": {
 			"type": "string",
 			"source": "html",
-			"selector": "h3"
+			"selector": "span"
 		}
 	},
 	"usesContext": [ "accordion/fontColor" ],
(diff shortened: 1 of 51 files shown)


ajvillegas (commented):
This looks and works great, but I'm not able to approve since I created the PR.

goldenapples (approved):

PR #233, opened 2025-07-25
1132: Add tabs block

joeleenk:
https://github.com/humanmade/Wikimedia/issues/1132

PR #246, opened 2025-08-22
1132: Add accordion item block deprecation

joeleenk:
While working on https://github.com/wikimedia/shiro-wordpress-theme/pull/233, I couldn't get the block deprecation working. But updating the block was really annoying, so I spent some more time and figured it out. 

This PR adds block deprecation for the accordion item block so the titles update.

### DIFF
--- a/assets/src/editor/blocks/accordion-item/deprecations.js
+++ b/assets/src/editor/blocks/accordion-item/deprecations.js
@@ -1,0 +1,43 @@
+/**
+ * Handle old versions of this block.
+ *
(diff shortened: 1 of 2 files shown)


goldenapples (approved):
The Save method still has what looks like an outdated function description in the docblock: https://github.com/wikimedia/shiro-wordpress-theme/blob/50688f0956fc312a8f8fce2553393f870fdf5cba/assets/src/editor/blocks/accordion-item/save.js#L5 

Otherwise, this looks good and should make updates much easier.

PR #317, opened 2026-03-12
Update header styles to fix z-index with jumplist

Sephsekla:
On annual reports using the jumplist, this block appears in front of megamenu dropdowns.

We can resolve this by increasing the z-index of the header so it will appear in front.


svandragt (approved):
Good until the jumplist takes 22!

